PR #1740, opened 2014-12-27
Support Ruby 2.2

dnrce:


matjack1:
I can't reproduce either :(

BTW, I've fixed the warnings in this commit: https://github.com/matjack1/tracks/commit/4b72d0fae997df0ddb29ba386e8854958e8e0898 but I cannot open a PR on your repo/branch. If there's no better way maybe just pick the modifications yourself.

dnrce:
The failures occur on Ubuntu but not in OS X.

matjack1:
cool I managed to reproduce the errors here: https://koding.com, that's a very good service :)

matjack1:
@dnrce I've found the problem and fixed in this commit: https://github.com/matjack1/tracks/commit/01511477b79daec7412a346969f6c4094d9411d2, if you want to cherry-pick it. Since there was the same name of active todos for the projects with id 2 and 3 the ordering was arbitrary.

Anyway you can understand the problem with this code change. Changing the fixtures was breaking lots of other tests, ending up in many changes, I don't know how do you prefer to continue.

dnrce:
Arbitrary, yes, but it should still be deterministic and shouldn't differ across platforms. We're missing something.

matjack1:
@dnrce I agree with you that the behaviour is unexpected, but try to do this in OSX and Ubuntu:
```
a = [[1, 2], [2, 4], [3, 4]]
a.sort{|x, y| y[1] <=> x[1]}
```
I get two different results. Is maybe an inconsistency of Ruby 2.2?

dnrce:
I asked on the #ruby-lang IRC channel and the best guess was that it's a difference in the underlying platforms' sort implementations. The real answer to ensure consistency is to make the sort condition deterministic, which I've done in 7f1246f42f91c6c0884d7c70be24f9511d40892d.

matjack1:
Interesting stuff and good fix ;)

PR #1743, opened 2014-12-30
Fix #1722

### LINKED ISSUE #1722: Recurrences (tests?) have an end-of-day edge case
See https://travis-ci.org/TracksApp/tracks/jobs/36919173 which ran at 2014-10-02T22:39:54Z.

```
Run options: --seed 4101

# Running:

..............................................................................................F.................................................................................................................................................................................F................................................................................F....................................................................................................................................................................................................................................................................

Finished in 65.736660s, 9.3403 runs/s, 25.9368 assertions/s.

  1) Failure:
RecurringTodos::DailyRepeatPatternTest#test_daily_every_day [/home/travis/build/TracksApp/tracks/test/models/recurring_todos/daily_repeat_pattern_test.rb:53]:
Expected: "02-10-14"
  Actual: "03-10-14"


  2) Failure:
UserTest#test_find_and_activate_deferred_todos_that_are_ready [/home/travis/build/TracksApp/tracks/test/models/user_test.rb:215]:
Expected: 0
  Actual: 1


  3) Failure:
TodoTest#test_load [/home/travis/build/TracksApp/tracks/test/models/todo_test.rb:25]:
Expected: "2014-09-26 00:00"
  Actual: "2014-09-25 00:00"

614 runs, 1705 assertions, 3 failures, 0 errors, 0 skips
```

From the Travis log, the actual start and stop times of the `rake ci` task were Thu, 02 Oct 2014 22:44:55 GMT and Thu, 02 Oct 2014 23:02:56 GMT, respectively.

matjack1:


matjack1:
@dnrce have you checked again this PR? I don't want to put pressure, I'm just making sure that everything is ok now

app/models/recurring_todos/abstract_repeat_pattern.rb
```diff
@@ -174,7 +174,7 @@ def continues_recurring?(previous)
     # same day as the previous
     def determine_start(previous, offset=0.day)
       start = self.start_from || NullTime.new
-      now = Time.zone.now
+      now = Time.now
```
dnrce: We should be handling time zones more fully, not removing zone usage. This change worries me.

test/models/todo_test.rb
```diff
@@ -22,8 +22,8 @@ def test_load
     assert_equal "Call Bill Gates to find out how much he makes per day", @not_completed1.description
     assert_nil @not_completed1.notes
     assert @not_completed1.completed? == false
-    assert_equal 1.week.ago.beginning_of_day.strftime("%Y-%m-%d %H:%M"), @not_completed1.created_at.beginning_of_day.strftime("%Y-%m-%d %H:%M")
-    assert_equal 2.week.from_now.beginning_of_day.strftime("%Y-%m-%d"), @not_completed1.due.strftime("%Y-%m-%d")
+    assert_equal Time.now.ago(1.week).beginning_of_day.strftime("%Y-%m-%d %H:%M"), @not_completed1.created_at.beginning_of_day.strftime("%Y-%m-%d %H:%M")
+    assert_equal Time.now.in(2.week).beginning_of_day.strftime("%Y-%m-%d"), @not_completed1.due.strftime("%Y-%m-%d")
```
dnrce: Same here. This is removing zone usage.

test/models/user_test.rb
```diff
@@ -207,7 +207,7 @@ def test_update_project_positions
   def test_find_and_activate_deferred_todos_that_are_ready
     assert_equal 1, @admin_user.deferred_todos.count
     @admin_user.deferred_todos[0].show_from = Time.now.utc - 5.seconds
-    @admin_user.deferred_todos[0].save
+    @admin_user.deferred_todos[0].save(:validate => false)
```
dnrce: Why this change?

app/models/recurring_todos/abstract_repeat_pattern.rb
matjack1: I'm not convinced by my fix either, so I'm happy to discuss about this :)
What I think is that in the DB we are saving all the dates without any TZ info. So I think that we should just translate to the relevant TZ when we display the data to the user.
From what I can see in this particular test, the date in the DB is saved in the UTC TZ, even if I start with the enviornment variable TZ set to something else. Is that a reliable way to do it? And if the data in the DB is without TZ info then we should compute everything accordingly, am I understanding right?

test/models/user_test.rb
matjack1: Since we are setting the show_from date to something in the past the validation was not passing, so it was not saved with the updated date and the test was not doing what was supposed to do.

app/models/recurring_todos/abstract_repeat_pattern.rb
matjack1: I'm looking at this commit: https://github.com/TracksApp/tracks/commit/3647d795875c0126d91e8151ee49e583bf2a174e maybe is relevant?

dnrce: Regardless of what the app tries to save in the database, ActiveRecord always converts to UTC before storing it, and will convert back on retrieval. Tracks should not need any of its own logic to handle this correctly.

matjack1: I see, so trying to leave the ```Time.zone.now``` in I experience a weird thing. When I launch the todo_test alone it works, when I launch the full suite it breaks.
If you try printing this in the ```test_load``` method:
```
p @not_completed1.created_at
p 1.week.ago.beginning_of_day
```
you will get two different results :confused:

It looks like the TZ is not loaded correctly when I launch a single test?
Is maybe a problem of libfaketime? Do you have any idea? Is maybe a clue to fix the problem?

dnrce: Maybe another test changes the timezone but doesn't revert it afterward?

matjack1: I've found the problem! It's the application_controller that sets the TZ based on the logged in user_preference, and then is not reverted.

I've added the site timezone as the default in the test helper and I've added a line in the fixtures in which we set the timezone again as the site default, so that the comparisons make sense, what do you think?

matjack1: @dnrce maybe you missed my last comment. Just pinging again to make sure that you see it, since I forgot to mention you before :)

dnrce:
This looks good to me. I have so little confidence in our time handling that it's tough to know whether green tests mean the behavior is actually correct. I'm cool with these changes though.

@matjack1, can you please rebase this?

matjack1:
here you go @dnrce ;)

PR #1746, opened 2015-01-20
remove unnecessary 'return'

HaiTo:
remove return if method body has 1 line.

i looking for..
app/**
lib/**

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -89,14 +89,14 @@ class ApplicationController < ActionController::Base
   end
 
   def count_deferred_todos(todos_parent)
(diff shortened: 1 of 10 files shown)


dnrce:
Thanks for the PR. I'm always torn on changes like this because while they definitely improve the code quality, they also add a layer of opacity to `git blame` that obscures history. Better to make the changes when the lines are otherwise touched. Leaving this one for someone else to decide.

C-Otto:
I believe such changes help the project and the PR should be accepted. Cleaning up code is part of development, and git is correc to "blame" the corresponding author. The effort needed to identify the cause of a potential bug affecting the same file is only increased in a very minimal way. I also think that clear code is more important than a clear git history.

HaiTo:
@dnrce, thank you for replay. I can understand that you do not want to change. 
so I remove 'return' if I send PR with another feature or bug fix.

thank you .

Issue #1455, opened 2014-06-25
CookieOverflow error on first-user signup


dnrce:
_Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/1455_

_Possibly_ related to #956?

Issue #956, opened 2014-06-25
Tracks crashes after login: CGI::Session::CookieStore::CookieOverflow


dnrce:
_Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/956_

After logging in Tracks crash and Thin returns code 500. A message from the logfile has been attached.

PR #1751, opened 2015-01-29
Fix ActionDispatch::Cookies::CookieOverflow after initial signup

br0ziliy:
Default :cookie_store can only hold 4K of data.
Upon first signup, following exception is thrown:
```
I, [2015-01-29T16:52:06.900449 #31028]  INFO -- : Started POST "/users" for 127.0.0.1 at 2015-01-29 16:52:06 +0100
I, [2015-01-29T16:52:06.945349 #31028]  INFO -- : Processing by UsersController#create as HTML
I, [2015-01-29T16:52:06.945514 #31028]  INFO -- :   Parameters: {"utf8"=>"✓", "authenticity_token"=>"8ihcH+0z8AZjZiCpgB9crHSpai/uxgaFDIZixdfEgzc=", "user"=>{"login"=>"vk password"=>"[FILTERED]", "password_confirmation"=>"[FILTERED]"}}
I, [2015-01-29T16:52:07.180296 #31028]  INFO -- : Redirected to http://localhost:3000/signup
I, [2015-01-29T16:52:07.180505 #31028]  INFO -- : Completed 302 Found in 235ms (ActiveRecord: 1.1ms)
F, [2015-01-29T16:52:07.182494 #31028] FATAL -- :
    ActionDispatch::Cookies::CookieOverflow (ActionDispatch::Cookies::CookieOverflow):
      actionpack (4.1.8) lib/action_dispatch/middleware/cookies.rb:529:in `[]='
      actionpack (4.1.8) lib/action_dispatch/middleware/session/cookie_store.rb:110:in `set_cookie'
...
```
This patch fixes this issue by changing session storage to the ActiveRecord session_store (for at least Ruby 2.0.0p353 / rails 4.1.8).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,8 @@ gem "cache_digests"
 # To use ActiveModel has_secure_password
 gem 'bcrypt', '~> 3.1.7'
 
+gem 'activerecord-session_store', github: 'rails/activerecord-session_store'
+
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 # gem 'turbolinks'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,12 @@
+GIT
+  remote: git://github.com/rails/activerecord-session_store.git
+  revision: 1cc0daf77d542045da4f6a43cba9ce04c27453dd
+  specs:
+    activerecord-session_store (0.1.1)
+      actionpack (>= 4.0.0, < 5)
+      activerecord (>= 4.0.0, < 5)
+      railties (>= 4.0.0, < 5)
+
 GEM
   remote: https://rubygems.org/
   specs:
@@ -211,6 +220,7 @@ DEPENDENCIES
   aasm
   actionpack-xml_parser (>= 1.0.1)
   activerecord-deprecated_finders
+  activerecord-session_store!
   acts_as_list
   aruba (>= 0.5.4)
   bcrypt (~> 3.1.7)
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -20,7 +20,7 @@ Rails.application.configure do
   # config.action_dispatch.rack_cache = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this).
-  config.serve_static_assets = true
+  config.serve_static_assets = false
 
   # Compress JavaScripts and CSS.
   config.assets.js_compressor = :uglifier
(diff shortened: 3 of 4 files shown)


dnrce:
Thanks for the contribution but I would prefer to stick with the CookieStore. The correct fix is to pare down what we are storing in the cookie -- there's *no way* it should be 4KB.

Please have a look at #956 and #1455 and see if what you're seeing seems related.

br0ziliy:
Hi,

I made couple of tests, and I think I know where is the root problem.
First, steps to reproduce:
- run below commands (*this will drop development database*):
```
bundle exec db:drop -e development
bundle exec db:migrate -e development
bundle exec rails server -e development
```
- go to http://localhost:3000/
- enter 2-letter username
- enter some password, and verification, and press Submit button
At this point the app should fail with CookieOverflow exception.
If it didn't - try using 3 or more letters for a user and it will fail for sure.

Now, where I think the problem is.
See below code in app/controllers/users_controller.rb:
```ruby
 81         user = User.new(user_params)
 82         
 83         unless user.valid?
 84           session['new_user'] = user
 85           redirect_to signup_path
 86           return
 87         end
```
On line 81 you create a new object, and then on line 84 you put the whole object in the session store.
Now, looking at the action_dispatch gem (lib/action_dispatch/middleware/cookies.rb), around the line 529:
```ruby
527         options[:value] = @encryptor.encrypt_and_sign(serialize(name, options[:value]))
```
After the above operation options[:value].size is around 4916 bytes in my case - which triggers the exception.

`options[:value]` before line 527 contain something like this:
```
{
"session_id"=>"1c05831fa0f021e853885a1c35bafec7", 
"return-to"=>"http://localhost:3000/", 
"expiry_time"=>2015-01-29 21:42:10 +0100, 
"_csrf_token"=>"sCC0iyJ1FoexUR+4ERFderT9G7Prk/P5/Y10Nl6wZlY=", 
"new_user"=>#<User id: nil, login: "vk", crypted_password: nil, token: nil, is_admin: false, first_name: nil, last_name: nil, auth_type: "database", open_id_url: nil, remember_token: nil, remember_token_expires_at: nil>
}
```
`serialize(name, options[:value])` (where `name` equals to the cookie name, `_trackapp_session`) produce a result of length 2617 bytes, and then these bytes are encrypted/signed, which doubles the size of what we want to store in a cookie.
I think that serialization/encryption of a User object that is stored in the session is the actual reason of the overflow exception.
Please let me know what do you think.

dnrce:
@br0ziliy, that seems believable -- and distinct from the older issues I mentioned before -- but I can't reproduce the error with these instructions. I'm going to close this pull request because it isn't the right fix, but can you please file a new ticket for this issue? Please also include the version of Tracks you're testing with, your browser version, etc. Thanks!!

PR #1758, opened 2015-02-09
repeating -> recurring, repeat -> recurrence

C-Otto:


dnrce:
You're correct that "repeating" becomes "recurring", but "repeat" should become "recurrence", as in `MonthlyRecurrencePatternTest`.

C-Otto:
Done.

dnrce:
Great. Nice cleanup!

lrbalt:
Hmm, too late again :-) 

There was a discussion on the mailing list at the time I added recurring todos where using repeat pattern / repeating todos was deemed better than recurring todos. I never got around to renaming it all to repeat pattern...

Anyway, better to be consistent :-)

PR #1759, opened 2015-02-09
occurence -> occurrence

C-Otto:
This just fixes a typo "occurence(s)" to "occurrence(s)", which also affects the database schema. I don't know if the "btree" entries really should be removed in the schema file. Please review.


dnrce:
I think if you re-run the migrations on MySQL you'll get the btree stuff back. That's the standard version we usually try to commit.

General note on commit messages: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

C-Otto:
So there's nothing I need to do?

dnrce:
Please run the migrations on MySQL to update schema.rb.

C-Otto:
Done.

dnrce:
Thanks! I get an empty diff when I run the migrations. Looks good.

PR #1762, opened 2015-02-16
Use supplemental bundler.d/*.rb files for easier packaging

GregSutcliffe:
This change allows easier packaging of the various components of Tracks' dependencies but using a .d-style Gemfile. This should be transparent to source-based installs, since all the same gems will be loaded by default, but for packaging, appropriate files can be left out (e.g. test.rb isn't useful in production) or packaged separately (e.g. tracks-mysql could contain mysql.rb). This allows a package-based installation to avoid unnecessary dependencies.

See https://groups.google.com/d/msg/tracksapp/ZEQSvav_puk/-PAYQRN7fg4J for background discussions on packaging.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,12 +16,6 @@ gem 'uglifier', '>=1.3.0'
 
 gem 'jquery-rails'
 
-# you may comment out the database driver you will not be using.
-# This will prevent a native build of the driver. Building native drivers is not
-# always possible on all hosters
-gem "sqlite3"
-gem "mysql2"
-
 gem "RedCloth"
 gem "sanitize"
 gem "will_paginate"
@@ -41,41 +35,7 @@ gem 'bcrypt', '~> 3.1.7'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 # gem 'jbuilder', '~> 1.2'
 
-group :development do
-  gem "spring"
-  gem "spring-commands-cucumber"
-  gem "yard"
-
-  gem 'tolk', '~> 1.6.0'
-
-  gem "bullet"
-  gem "rack-mini-profiler"
-  gem "rack-dev-mark"
-  gem "activerecord-deprecated_finders"
-end
-
-group :test do
-  gem "factory_girl_rails"
-  gem "capybara"
-  gem "cucumber-rails", :require => false
-  gem "rspec-expectations"
-  gem "database_cleaner"
-  gem "mocha", :require => false
-
-  gem "aruba", ">=0.5.4", :require => false
-
-  # Note that > 2.14 has problems, see:
-  # https://code.google.com/p/selenium/issues/detail?id=3075
-  gem "selenium-webdriver"
-
-  # uncomment to use the webkit option. This depends on Qt being installed
-  # gem "capybara-webkit"
-
-  # uncomment to be able to make screenshots from scenarios
-  #gem "capybara-screenshot"
-  #gem "launchy"
-
-  gem "simplecov"
-  # get test coverage info on codeclimate
-  gem "codeclimate-test-reporter", group: :test, require: nil
+# Load supplemental groups from bundler.d
+Dir["#{File.dirname(__FILE__)}/bundler.d/*.rb"].each do |bundle|
+  self.instance_eval(Bundler.read_file(bundle))
 end
--- a/bundler.d/development.rb
+++ b/bundler.d/development.rb
@@ -1,0 +1,12 @@
+group :development do
+  gem "spring"
+  gem "spring-commands-cucumber"
+  gem "yard"
+
+  gem 'tolk', '~> 1.6.0'
+
+  gem "bullet"
+  gem "rack-mini-profiler"
+  gem "rack-dev-mark"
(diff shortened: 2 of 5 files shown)


lrbalt:
I have no objections to the changes of Gemfile. I would like to see updates for the documentation though, since people need to run bundler in a different way, both for production and for development/test. Do we need to update the Travis.yml too?

GregSutcliffe:
I *think* Travis won't notice a difference, since all the code to load the gems happens within bundler itself. I'll try to run the tests and check. 

Happy to add docs - in general, people don't need to change behaviour - using `--without <group>` will still work correctly, since the groups are still specified in their respective files. People wishing to package Tracks may wish to be aware of the option to not package certain files though - it's *far* easier to not package a file than to get your `bundler --without ...` command correct for everyy possible circumstance.

Is there a good place to add docs? I'm happy to add a note to the Gemfile itself, anything else?

dnrce:
I'm not a fan of deviating from normal Gemfile layout, just because one of Tracks' (unofficial? or at least my own) goals is to demonstrate good Rails practices. Can you elaborate more on why this is of benefit for packaging? I'm curious how your Debian package build would be improved by splitting these files.

GregSutcliffe:
@dnrce the idea is that we don't package things like development.rb, test.rb, and even the database files in the core Tracks package. This allows me to produce a package with minimal dependencies without having to sed or otherwise modify the Gemfile (keeping my own Gemfile and staying in sync is a nightmare, particularly for automated nightly builds).

I can then produce packages like tracks-mysql, tracks-sqlite, and tracks-test which simply lay down the additonal files into bundler.d and call a bundle update. For an example, I can point you to https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/wheezy/foreman/foreman-test.install and https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/wheezy/foreman/foreman-test.postinst which do the same thing there.

Does that help? Can I elaborate elsewhere? How are we breaking Rails practices here (since effectively it's still all controlled from Bundler?)

dnrce:
I guess I'm still unclear how this is an improvement over the current (or even expanded) use of native bundler groups. What's hard about getting the `--without` arguments right?

lrbalt:
If we go this route, we need to explain in the installation and upgrade guide in /doc how to install Tracks manually using --without to remove sqlite or mysql group. Or do the default instructions work with this change?

Do we need to use a --without in travis.yml?

GregSutcliffe:
@lrbalt current instructions continue to work - Bundler will load every file in the directory, so the end result remains the same (you'll see the PR doesn't contain changes to Gemfile.lock, and I did run a bundle install with this change before my git commit, so all the same gems are present). Adding db groups and documenting the option isn't a bad idea even if this isn't merged though. I can do another PR for that if need be (see below).

@dnrce good question (and apologies for the wall-o-text thats incoming). If it was just the development and test groups (i.e. things a package-based user wouldn't normally care about), I'd agree with `--without` being usable.

The issue comes when you want to support multiple optional components without just being heavy-handed and installing all the deps, and is related to bundler using `without` instead of `with` (which would be *much* easier!). Suppose I create a core package called "tracks" without and db deps (so, `bundler --without mysql --without sqlite`). I now create two smaller packages that enable extra functionality (so, "tracks-mysql" which calls `bundler --without mysql`, and vice-versa for "tracks-mysql"). Even this works fine, until you want more than one db package (maybe you're doing a migration...). Once you install both, you have fundamentally conflicting bundler calls - only the last one to install will win. You're left with doing something like using `sed` on the bunder config to achieve your aim of giving the user maximum flexibility, which sounds more fragile to me than this approach.

In fairness, this approach comes into it's own once you have a lot of optional compnents (currently Foreman, which I stole this approach from, has ~12, some of which have large dependancy lists. It also includes separating out gems for asset compilation, which is very handy). It's also handy if you end up with plugins (i.e. Rails engines) being packaged, as it's just another file in bundler.d.

In relation to Tracks, which right now really only has the db gems as optional-and-simultaneous things, I can see how the benefit is less clear. There's probably only 2 gems and 2-3 deb packages being installed (although I would like to separate out the asset deps as well, which would cut back a lot). It doesn mean we have less packages to build if we want to try a non-vendored (i.e. one package per gem) approach though. 

If we don't go this way, can I at least send another PR to add groups around the db gems? I'm actually using `sed` to remove mysql at the moment ([see here](https://github.com/GregSutcliffe/tracks_packaging/blob/master/debian/rules#L12)) which is ugly as all hell :)

GregSutcliffe:
Oops, missed one extra comment - with respect to paragraph 3, consider also the multiplicative effect of adding new groups to upstream. The downstream "tracks" package needs to call `bundler --without` for every group, since (I don't think) there's a `bundler --without all-groups`. Every time upstream adds a group, the packaging needs rebuilding, making the combinatiorial problem worse.

GregSutcliffe:
Seems we're struggling to find consensus on this, and I have to admit, it has less benefit here than I first thought. I'm happy to close this PR for now, but I'll probably send a smaller one for making database bundler groups, just so I don't have to sed the Gemfile ;)

PR #1763, opened 2015-02-19
Use the Rails helpers in more places

mattr-:
I was reading through TodosHelper and noticed a few places where we weren't using the helpers Rails provides for thinks like model ids and routes, so I fixed those up. I also added some very rudimentary characterization tests along the way.


dnrce:
Looks great!

PR #1764, opened 2015-02-21
Add db bundler groups so they can be excluded with --without

GregSutcliffe:
Simple groups so that the packaging can split out dependencies appropriately

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,11 +16,12 @@ gem 'uglifier', '>=1.3.0'
 
 gem 'jquery-rails'
 
-# you may comment out the database driver you will not be using.
+# you may comment out the database driver(s) you will not be using.
 # This will prevent a native build of the driver. Building native drivers is not
-# always possible on all hosters
-gem "sqlite3"
-gem "mysql2"
+# always possible on all platforms
+# Alternatively use --without <group> arguments to bundler to not install that group
+gem "sqlite3", group: :sqlite
+gem "mysql2", group: :mysql
 
 gem "RedCloth"
 gem "sanitize"


dnrce:
I like the approach. Seems very clean. This is the same strategy used by GitLab: https://github.com/gitlabhq/gitlabhq/blob/master/doc/install/installation.md#install-gems

Actually, I really like their one-liner syntax. Can you update this PR to match?

GregSutcliffe:
Sure, I can do that. Their style is Ruby 1.9 only (foo: :bar rather than :foo => :bar) - is Tracks 1.9+ at this point anyway? Or shall I use the older style?

Also, any objection to adding Postgresql? I know it works for me, but I'm not sure if you want to make it official ;)

dnrce:
Tracks 2.3+ has dropped Ruby 1.8 support, so the new syntax is fine.

I think adding Postgres is a good idea, particularly to aid the use of Tracks on Heroku. But it comes with some additional considerations that might warrant discussion (developers will need to install additional system libraries, schema.rb format differs, etc.) so it should be a separate change.

GregSutcliffe:
That's fine, pushed the new syntax. I'll test that at least the migrations work on psql and push a separate PR for that later.

GregSutcliffe:
Oh, changed the comment wording slightly, too. It's somewhat in prep for psql but reads fine even without.

dnrce:
:shipit:

dnrce:
This does not cause a behavior change with the current install procedure, so I went ahead and merged it, but the install docs would benefit from an update.

GregSutcliffe:
Thanks! Opened https://github.com/TracksApp/tracks/pull/1765 to address the docs. I'll try to remember to update that further whe ndoing the psql PR.

PR #1765, opened 2015-02-21
Update install doc to use db groups

GregSutcliffe:
As requested following the db groups Gemfile update.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -50,8 +50,20 @@ You need to create a database and database-user to use with Tracks. For this, yo
 
 Tracks is built upon a number of Ruby libraries (known as ‘gems’). The Bundler tool makes it easy to install all the gems that Tracks needs, and ensures that they are all the correct versions.
 
-1. Edit the file `Gemfile` in the Tracks root directory. You may comment out any database drivers you will not be using. You may also comment out `therubyracer` if you are installing Tracks on Windows or Mac OS X, or if you have another JavaScript runtime such as Node.js installed.
-2. Run the command `bundle install --without development test` in the directory that you unzipped your Tracks download to.
+1. If you are installing Tracks on Windows or Mac OS X, or if you have another JavaScript runtime such as Node.js installed, edit the file `Gemfile` in the Tracks root directory, and comment out `therubyracer`.
+2. In the directory that you unzipped your Tracks download to, install all the appropriate gems by running:
+  * If you are using SQLite:
+
+        ```
+        bundle install --without development test mysql
+        ```
+
+  * If you are using MySQL:
+
+        ```
+        bundle install --without development test sqlite
+        ```
+
 3. Wait for Bundler to finish installing the necessary gems that Tracks needs. This can take some time depending on the speed of your internet connection and the speed of the system you’re installing Tracks on.
 
 ## Configure variables


GregSutcliffe:
This makes the bundler section look slightly different, but I wanted to give users lines they could cut'n'paste. Asking a user to remember which db they're *not* using and then putting it on the `--without` line in a later step is likely to cause confusion otherwise.

dnrce:
I like the sentiment, but this ends up being way too complicated. Let's follow GitLab's example instead: https://github.com/gitlabhq/gitlabhq/blob/master/doc/install/installation.md#install-gems

GregSutcliffe:
Gah, I'm clearly more tired than I thought. That's far better, thanks. Updated.

doc/installation.md
```diff
@@ -50,9 +50,15 @@ You need to create a database and database-user to use with Tracks. For this, yo
 
 Tracks is built upon a number of Ruby libraries (known as ‘gems’). The Bundler tool makes it easy to install all the gems that Tracks needs, and ensures that they are all the correct versions.
 
-1. Edit the file `Gemfile` in the Tracks root directory. You may comment out any database drivers you will not be using. You may also comment out `therubyracer` if you are installing Tracks on Windows or Mac OS X, or if you have another JavaScript runtime such as Node.js installed.
-2. Run the command `bundle install --without development test` in the directory that you unzipped your Tracks download to.
-3. Wait for Bundler to finish installing the necessary gems that Tracks needs. This can take some time depending on the speed of your internet connection and the speed of the system you’re installing Tracks on.
+1. If you are installing Tracks on Windows or Mac OS X, or if you have another JavaScript runtime such as Node.js installed, edit the file `Gemfile` in the Tracks root directory, and  comment out `therubyracer`.
```
dnrce: extra space before "comment"

doc/installation.md
```diff
@@ -50,9 +50,15 @@ You need to create a database and database-user to use with Tracks. For this, yo
...
+1. In the directory that you unzipped your Tracks download to, install all the appropriate gems by running:
+
+        # For SQLite users:
+        bundle install --without development test mysql
+        # For MySQL users:
+        bundle install --without development test sqlite
+
+1. Wait for Bundler to finish installing the necessary gems that Tracks needs. This can take some time depending on the speed of your internet connection and the speed of the system you’re installing Tracks on.
```
dnrce: Please keep the actual numbering -- it is a little tedious when making changes, but it keeps the file more readable as plaintext for users who are opening the file from within their downloaded copies of Tracks.

doc/installation.md
```diff
@@ -50,9 +50,15 @@ You need to create a database and database-user to use with Tracks. For this, yo
 
 Tracks is built upon a number of Ruby libraries (known as ‘gems’). The Bundler tool makes it easy to install all the gems that Tracks needs, and ensures that they are all the correct versions.
 
-1. Edit the file `Gemfile` in the Tracks root directory. You may comment out any database drivers you will not be using. You may also comment out `therubyracer` if you are installing Tracks on Windows or Mac OS X, or if you have another JavaScript runtime such as Node.js installed.
-2. Run the command `bundle install --without development test` in the directory that you unzipped your Tracks download to.
-3. Wait for Bundler to finish installing the necessary gems that Tracks needs. This can take some time depending on the speed of your internet connection and the speed of the system you’re installing Tracks on.
+1. If you are installing Tracks on Windows or Mac OS X, or if you have another JavaScript runtime such as Node.js installed, edit the file `Gemfile` in the Tracks root directory, and  comment out `therubyracer`.
+1. In the directory that you unzipped your Tracks download to, install all the appropriate gems by running:
+
+        # For SQLite users:
```
dnrce: Can you make these sub-bullets, with one line of code each? Being a single code block makes it a little less clear that it's one or the other.

Also, it should be "If you will be using SQLite", not "For SQLite users" -- they aren't configuring for third parties.

doc/installation.md
GregSutcliffe: Fixed, thanks.

doc/installation.md
GregSutcliffe: Good point, fixed.

doc/installation.md
GregSutcliffe: This turns out to be surprisingly hard to do - I spent about 20mins on different GFM syntax attempts, none of which worked the way you suggest. Code blocks in lists have to be indented by 8 spaces, but that seems to mess with sublists - simply indenting the code by 8 or 10 spaces after a sublist  entry doesn't produce the correct result.

I'm open to ideas, but for now the best I can do is separate it into 2 code blocks, which seems to work well, but may not be visually distinct enough.

dnrce:
@GregSutcliffe, check out https://gist.githubusercontent.com/clintel/1155906/raw/78c0ddea05608ef082343911c3de852bebf30e9f/gistfile1.md for syntax.

doc/installation.md
dnrce: Check out https://gist.githubusercontent.com/clintel/1155906/raw/78c0ddea05608ef082343911c3de852bebf30e9f/gistfile1.md for syntax.

GregSutcliffe: Ah, don't you just love GFM. Thanks, updated.

dnrce:
Looks good now. Thanks!

dnrce:
And a general note -- it's preferable to open PRs from feature branches, not your `master`.

GregSutcliffe:
Surely the PR system makes it irrelevant wher I choose to open it from? It'll have zero effect on your side of the repo. Of course, if you don't like seeing ".../master" in the merge commit, then use --ff-only to remove merge commits (which also makes your git history *much* less messy :) )

GregSutcliffe:
Oh, and thanks! :)

dnrce:
Using your `master` has a few negative ramifications on both sides. GitHub itself, and many others, recommend using topic branches -- it's an easy thing to Google. You're right, it isn't _necessary_, but it prevents some easily avoidable annoyances.

GregSutcliffe:
I should probably be more clear - I generally do use feature branches for complex entities which will take a while to review, and may need rebasing along the way. Small fixes like this though, I generally don't bother.

I really wish GitHib would default to using fast-forward merges, it would save an *awful* lot of churn in the logs. I get around it by merging on the commandline and pushing the result back up, but it'd *so* much easier to use the big green button in the UI.

I can use feature branches if you wish, I was more curious about the reasoning behind the request (and clearly, I made some assumptions too).

PR #1766, opened 2015-02-22
CI test on both MySQL and SQLite

dnrce:
This makes SQLite fully supported, at the cost of significantly longer CI build times.


dnrce:
This could be limited to model tests only, but our MVC separation isn't currently clean enough.

PR #1767, opened 2015-02-25
Place therubyracer in a gem group

dnrce:
With this change and @GregSutcliffe's previous work in #1764 and #1765, users should never need to edit their Gemfiles to install Tracks.


GregSutcliffe:
:+1: reads fine to me.

mattr-:
A toast to you my good sir.

leo-toast

PR #1769, opened 2015-02-25
Use Thin instead of WEBrick

dnrce:
Thin performs much better than WEBrick, and it's a drop-in replacement, requiring no operational or development changes.


lrbalt:
This could be helpfull if we want to use streaming updates one day

PR #1770, opened 2015-02-25
Update Gemfile and install doc to use db groups

GregSutcliffe:
This is a backport from master of #1764, #1765, and #1767

One of the aims of was to make it easier for me to package Tracks, and I'm currently packaging the 2.3 branch. So, having verified that the groups approach is acceptable (by getting it into master), here is the 2.3 backport. I've checked that bundle install still works and the versions are correct:

```
Installing sqlite3 1.3.10
Installing mysql2 0.3.16
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem 'coffee-rails', '~> 4.0.1'
 gem 'actionpack-xml_parser', '~> 1.0.1'
 
 # See https://github.com/sstephenson/execjs#readme for more supported runtimes
-gem 'therubyracer'
+gem 'therubyracer', group: :therubyracer
 
 gem 'uglifier', '~> 2.5.3'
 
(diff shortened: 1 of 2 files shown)


GregSutcliffe:
Bump. I don't really want to have to build the mysql-capable packages from my 2.3 branch.

dnrce:
Being in release candidate phase means we're bugfix-only until release. I'll leave this PR open, but it should not be merged at the present time.

GregSutcliffe:
That's going to cause me an issue for building packages (without evil hacks) for this very release, which my intention all along. This doesn't touch any code or change any gem versions, as demonstrated. Are there other concerns that could be addressed?

GregSutcliffe:
Actually, in your defense, that's not strictly true. I can make the package depend on both database packages (libsqlite3-dev and libmysqlclient-dev). However, it's slightly more user-friendly to not have to do that - and more user-friendly installs are definitely something I'd like to work towards.

mattr-:
:+1: :shipit:

This will at least make it into 2.3.1 :smiley:

GregSutcliffe:
@mattr- Thanks! I've got new packages slated for sometime soon, just need to update the toolchain for Jessie. This will help ;)

PR #1777, opened 2015-03-15
fix #1754

### LINKED ISSUE #1754: Cyclic dependency
I created a few new tasks in a fresh project and then connected those using drag-n-drop. By accident, I created a dependency of taskA to taskB, although I created the reverse dependency just before that. I remember seeing some red error message flash up (which I ignored/did not read). I assumed the dependency was not added, as it would create a cycle.

However, according to my database, there is the cyclic dependency:

```
tracks=# select * from dependencies where successor_id in (5063,5064);
  id  | successor_id | predecessor_id | relationship_type 
------+--------------+----------------+-------------------
 1098 |         5064 |           5063 | 
 1099 |         5063 |           5064 | 
```

Running with PostgreSQL and on f81efebd8ce18bbca9abd567d345edbc22e80061.

C-Otto:
Move the validation to 'Dependencies', fix test case

### DIFF
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -257,9 +257,13 @@ class TodosController < ApplicationController
     @todo = current_user.todos.includes(Todo::DEFAULT_INCLUDES).find(params['successor'])
     @original_state = @todo.state
     unless @predecessor.completed?
-      @todo.add_predecessor(@predecessor)
-      @todo.block! unless @todo.pending?
-      @saved = @todo.save
+      begin
+        @todo.add_predecessor(@predecessor)
+        @todo.block! unless @todo.pending?
+        @saved = @todo.save
+      rescue ActiveRecord::RecordInvalid
+        @saved = false
+      end
 
       @status_message = t('todos.added_dependency', :dependency => @predecessor.description)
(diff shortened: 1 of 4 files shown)


C-Otto:
This changes the error message, as the GUI code only checks if the task can be saved. The exception now thrown is not handled.

C-Otto:
Now the error message should be the same (at least in my manual tests), but I'm not too happy with the code. If you manage to create a cleaner version, feel free to enlighten me :)

C-Otto:
When trying to create the cyclic dependency, the todo is removed from the view. Instead, when the error message is shown, the UI should stay as it was before.

lrbalt:
thanks!

dnrce:
This should have been done against 2.3_branch then merged down to master. I'll cherry-pick it.

C-Otto:
I think the UI feedback is still wrong, as indicated in my last comment.

dnrce:
Do you think this should not have been merged?

C-Otto:
I'll check (today).

C-Otto:
When using drag and drop, the dragged action (closing the cycle if added as a successor) disappears.
When using the 'edit' menu, no error is shown, and the popup cannot be closed using 'Update'.

PR #1783, opened 2015-03-21
do not activate already completed todos, fix #1778

### LINKED ISSUE #1778: Showing completed tasks due to dependencies
The Situation:

B depends on A. First, B is marked as completed, although A still needs to be done. Then A is marked as completed.

Actual result:

A disappears, B is shown as a new todo. Marking B as done shows an error. Marking B as done again makes it disappear. I guess the reason is that dependent todos are shown without checking their completion status.

Expected result:

I guess this is up for discussion, as I started with a situation that "should not happen". Although, it happened more than once to me already, as I sometimes like to introduce dependencies to manage my tasks, instead of hiding really dependent tasks (first work than play vs. first find woman then father a child).

C-Otto:
Only activate successors which are not completed

### DIFF
--- a/app/models/todo.rb
+++ b/app/models/todo.rb
@@ -317,7 +317,7 @@ class Todo < ActiveRecord::Base
 
   # activate todos that should be activated if the current todo is completed
   def activate_pending_todos
-    pending_todos = successors.select {|t| t.uncompleted_predecessors.empty?}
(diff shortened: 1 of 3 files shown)


dnrce:
Thanks! Can you please include a test?

C-Otto:
Done.

PR #1787, opened 2015-03-31
Cucumber fix

C-Otto:
'Scenario: Deleting a todo will remove it from the calendar' fails on the last day of the month

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ group :test do
 
   gem "aruba", ">=0.5.4", :require => false
 
-  gem "selenium-webdriver"
+  gem "selenium-webdriver", '>=2.45'
 
   # uncomment to use the webkit option. This depends on Qt being installed
   # gem "capybara-webkit"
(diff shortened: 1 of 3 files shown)


GregSutcliffe:
@C-Otto there's no requirement for 2.14 in the codebase, and 2.43 is listed in the Gemfile.lock (https://github.com/TracksApp/tracks/blob/master/Gemfile.lock#L157) so I can't see why you'd be getting 2.14 unless it was in your lockfile from some time ago. I agree the comment is outdated though.

I'll try and test the actual fix in the next day or two ;)

C-Otto:
Version 2.14 was just a version mentioned in a comment (outdated for three years). Version 2.43 is too hold, which is why I demand 2.45 in this commit.

GregSutcliffe:
Right, that makes more sense - I think there's just miscommunication over wording here. If you want to update to 2.45 you probably want to add an updated Gemfile.lock to the commit, otherwise `bundle install` will not be able to satisfy the Gemfile with the Gemfile.lock

C-Otto:
I will close this and create new pull requests for the issues.

PR #1801, opened 2015-04-04
replace 'None' project by empty string

### LINKED ISSUE #440: 'Project' field in action's edit mode should be highlighted
*Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/440*

Once in a while (due to my indecisive nature I guess) when I create an action, I do not associate a project with it. It could also be that I forgot to create the project before creating the action.

Later, when I go to the edit mode of the action (in the front page), I see that "None" is filled in the Project section. Changing this becomes a pain - I need to select the whole text and then delete it and then enter a new one. Couldn't Tracks do this instead:

In the edit mode of an action, whenever the curson/mouse is in the Context or the Project field:

- If the value is "None", the whole value is deleted

- If the value is anything else, the value is highlighted/selected

*Originally reported by Anonymous on February 2, 2007 at 01:33:28 (-0800)*

C-Otto:
I don't see why 'None' is better than the empty string, and this fixes #440. If you prefer still having 'None', please tell me what the advantages are.

Fix #440

### DIFF
--- a/app/controllers/todos/todo_create_params_helper.rb
+++ b/app/controllers/todos/todo_create_params_helper.rb
@@ -104,7 +104,6 @@ module Todos
     def project_specified_by_name?
       return false if @attributes['project_id'].present?
       return false if project_name.blank?
-      return false if project_name == 'None'
       true
     end
 
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -1135,7 +1135,7 @@ end
   def update_project
     @project_changed = false;
     if params['todo']['project_id'].blank? && !params['project_name'].nil?
-      if params['project_name'] == 'None'
+      if params['project_name'].blank?
         project = Project.null_object
       else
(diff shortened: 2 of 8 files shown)


lrbalt:
None is used in the mobile interface IIRC. I'd like not to change this, especially in RC-phase

C-Otto:
I checked for usages, it is not used.

Could you create a 2.4 (RC) branch? This would help working on features not intended for 2.4 (not just this one).

dnrce:
@lrbalt, this PR is against master so it has no bearing on the 2.3 release candidate. We should only be merging 2.3 to master at this point, not the other way around.

lrbalt:
@dnrce Sorry, you are right. I'm out of the game to long :-) I'll fix my own mistakes and cherry-pick to get both trees in sync again. I'm not merging this one yet, I want to look into the use of None

dnrce:
Great change -- this one has bugged me. Thanks!

PR #1807, opened 2015-04-10
Use standard CSS for rounded corners

dnrce:
Remove Nifty Corners and prefixed CSS rules.
Browser support of CSS3 border-radius is excellent by now.

http://caniuse.com/#feat=border-radius


mattr-:
Yay! :tada:

dnrce:
Turns out not all of our corners are already CSS rounded; some are still Nifty Corners only -- this isn't ready to merge until I address that.

dnrce:
Added CSS rules to replace Nifty-Corners-only instances.

PR #1808, opened 2015-04-11
Projects with completed actions may be blocked

C-Otto:
Currently, a project cannot be blocked if at least one completed action exists. This was introduced in
acab98d4c7f2237bcd9fc44cd5e93a39582fe588, and I assume this change was not intended.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -109,7 +109,7 @@ class Project < ActiveRecord::Base
     ## mutually exclusive for stalled and blocked
     # blocked is uncompleted project with deferred or pending todos, but no next actions
     return false if self.completed?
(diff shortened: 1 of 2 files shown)


lrbalt:
I need to look into this. I've been out of the game for a long time, so I need to check the semantics. I remember making this change for a reasong

C-Otto:
The question is "Should a project with at least one completed action possibly be marked as blocked?". If the answer is yes, my commit solves the issue.

dnrce:
@lrbalt, can you please take a look here and see if you can remember? This is a change from 2.2.

dnrce:
@lrbalt, ping

lrbalt:
AFAICS, the change is correct. @dnrce you can merge if you like

PR #1809, opened 2015-04-12
Freeze time in tests

### LINKED ISSUE #1760: Intermittent StatsController test failure
See https://travis-ci.org/TracksApp/tracks/jobs/50132324, which built 52481e0 at 2015-02-09 23:48:31Z.

```
1) Failure:
StatsControllerTest#test_actions_open_per_week_data [/home/travis/build/TracksApp/tracks/test/controllers/stats_controller_test.rb:252]:
there are actions in the first 17 weeks of this year.
Expected: 17
  Actual: 18
```

C-Otto:
fix #1760 by not advancing the time during the tests

### DIFF
--- a/test/models/todo_test.rb
+++ b/test/models/todo_test.rb
@@ -21,7 +21,7 @@ class TodoTest < ActiveSupport::TestCase
     assert_equal 2, @not_completed1.project_id
     assert_equal "Call Bill Gates to find out how much he makes per day", @not_completed1.description
     assert_nil @not_completed1.notes
-    assert @not_completed1.completed? == false
+    refute @not_completed1.completed?
     assert_equal 1.week.ago.beginning_of_day.strftime("%Y-%m-%d %H:%M"), @not_completed1.created_at.beginning_of_day.strftime("%Y-%m-%d %H:%M")
     assert_equal 2.week.from_now.beginning_of_day.strftime("%Y-%m-%d"), @not_completed1.due.strftime("%Y-%m-%d")
     assert_nil @not_completed1.completed_at
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -15,6 +15,7 @@ class ActiveSupport::TestCase
   def setup
     Time.zone = SITE_CONFIG['time_zone']
     @today = Time.zone.now
+    travel_to @today
     @tomorrow = @today + 1.day
     @in_three_days = @today + 3.days
     @in_four_days = @in_three_days + 1.day    # need a day after start_from
@@ -26,6 +27,10 @@ class ActiveSupport::TestCase
     @tuesday = Time.zone.local(2008,6,10)
     @wednesday = Time.zone.local(2008,6,11)
     @thursday = Time.zone.local(2008,6,12)
+  end
+
+  def teardown
+    travel_back
   end
 
   # Add more helper methods to be used by all tests here...


C-Otto:
The time reported by `Time.now` is not advanced, no matter how much time it takes to actually run the tests. So even if the actual time passes over to the next day, the time reported inside the tests is still the old time.

I do not see any need to sleep inside the tests, and I do not see the dependency to have time advance. Could you please have another look into this?

I replaced `assert !x` by `refute x`.

dnrce:
I don't like this. We should not be fixing the time in our tests. Tracks needs to behave well all the time, and we have caught a number of bugs by running the tests using real times. If particular tests have deficiencies, they should be addressed directly.

C-Otto:
The real time is still used, it is just not advanced. It seems you're confusing the current time with the time difference between the start of the test, and some code inside the test.

I agree that the current time should be used, and not some fixed date. This helps finding bugs only visible on February 29th, or only on the last day of a month, etc. (assuming the tests are run then).

But I do not think that time needs to advance while a test runs. This just simulates an infinitely fast computer (assuming no `sleep` is used). Do you really think the time needed to execute tests helped catching bugs?

dnrce:
I wasn't confusing the two but I did not word my last message well. I think it's not good testing practice when it can be avoided. I also think that for the few cases where time needs to be frozen, the invocation should be placed very close to the test so that the context is more clear. I do not like global freezing.

That being said, I looked at the actual tests that are failing and frankly, they're pretty bad tests and it stems from our use of a large set of fixtures. Freezing time might be the best *temporary* patch to the tests until they can be improved. So I'll reopen and let @lrbalt weigh in on this.

dnrce:
It would be helpful to add some reasoning to your commit message. For example, mention the fact that the fixtures are all loaded at test setup time.

Also, the false/refute change has no relation to the rest of this PR and is confusing to include here.

C-Otto:
I see several tests depending on what exactly 'one week ago' or so means, where the subtle time difference between creating the actions and checking for their existence is crucial when the test is started around midnight (or maybe on the last day of a month).

This is not a problem of badly written tests nor fixtures, I think. I in fact, the following test also has this problem:

```
now = Time.zone.now
tomorrow_time = today + 1.day

today = Date.new(now.year, now.month, now.day)
tomorrow = Date.new(tomorrow_time.year, tomorrow_time.month, tomorrow_time.day)

refute today === tomorrow
```

This test may fail if enough time passes between the first two lines. If we want to use the 'now' time, and we both think we should, I don't see another way of working around the issue of 'now' changing while the test runs.

I'll try to leave out unrelated commits (although this means some minor PRs just for those).

dnrce:
(I removed my comment about the unrelated commit. One or two "cleanup" commits as part of a PR are fine, _if_ there is only one actual "meat" commit. Where it gets confusing is if the PR includes multiple meaningful commits and one or two meaningless ones.)

dnrce:
That example *is* a bad test. If the second line is rewritten as `tomorrow_time = now + 1.day`, there will be no problem.

C-Otto:
I think the tests cannot be better, because it depends on how time is used in production code. Outside of tests, the data was entered by the user before the code is run. It is natural to use 'Time.zone.now' inside the code. However, inside tests, the exact value of 'Time.zone.now' as used during the test execution cannot be known in advance, which causes the mentioned problems.

The production code should not be changed, as working with the current time is exactly right. However, the test can modify the world the production code is running in, which is what my commit does.

Please think about how a good test look like. In the (wrong and removed) example you're mentioning, you'd need to rewrite a line which actually is part of the production code. Rewriting anything in the test itself, preparing the data and checking the results, does not solve the issue at all.

dnrce:
The main problem with the tests is that the fixtures are loaded at the beginning of the test suite, whereas the tests themselves assume that fixture data was created at the time the individual test is run. To me, that's the only valid excuse for freezing time globally. Other specific cases should be addressed more directly and more granularly.

Application code should not use `Time.now` (actually, `Time.current` is better, but that's another story) more than once within a method, either. If multiple relative times are needed, the method should store a temporary time object and create other times relative to it. If we're not doing that, it's a bug.

C-Otto:
It does not make a difference when exactly the fixtures are loaded, as the application code runs strictly after it, no matter what. Even a single nanosecond of difference between loading the fixtures and running the tested application code can cause this kind of bug.

Again, the issue is not what the tests assume to be true. The issue is that time passes inside the application code. You can't work around that, no matter what you do in the test code.

Even if the current time is only retrieved once inside the application code, this may still cause this kind of bug. Let me try to write another example illustrating this, this time using pseudo code.

```
prepare_test_data(today, 'some test data')
response = run_application_code # which does just return load_from_database(today)
assert_equal response, 'some test data'
```

Assuming the application code under test just returns the data found for the current day, you will run into problems if the test is started just a short bit before midnight. If the day changes between the first and second line, the application code won't find the prepared data, and the test will fail.

How would you solve this? I assume you don't want to change the application code?

dnrce:
We have been talking past each other here a little bit. I totally agree that there isn't a good way to solve that example without freezing time, and using factories instead of fixtures wouldn't help. What using factories _would_ do is let you freeze time for _only that one test_, because the factory calls are within the test. That's the limited scope of time-freezing that I would prefer to see.

C-Otto:
Ah, I understand. However, I don't think it would be easy to identify the tests that need this kind of fix (without waiting for bugs to pop up).

dnrce:
Agreed. It's also an issue of our test scopes overlapping too much -- our integration and acceptance tests go way too deep into what should be covered by unit tests. But I sure won't give this PR a hard time for that. :-)

lrbalt:
I have fixed some tests. The problem with Tracks is not time progressing a few microseconds, but the timezone calculation and the use of to_midnight. This causes 00:00 +0100 to be converted to 23:00 +0000. This messes up comparison:  a_date.utc.at_midnight is not the same as a_date.at_midnight. We use at_midnight in several places where the timezone could be different. 

I'd like to close this pull request and not fix time, since it will not fix any potential problems but will prevent them from showing.

C-Otto:
Fine with me, I must admit I don't really understand the time computations in tracks.

lrbalt:
I often have that same feeling :-)

PR #1812, opened 2015-04-12
Circular dependencies

C-Otto:
Fix two issues already mentioned in pull request #1777. This should be added, as #1777 is imcomplete/buggy without. Also relevant for release of 2.3.

1) Drag-n-drop removes the dragged todo if adding the dependency fails.
2) Adding a dependency does not show any error message in the edit form.

Also relevant for #1754, obviously.

### DIFF
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -423,7 +423,16 @@ class TodosController < ApplicationController
     update_dependencies
     update_attributes_of_todo
 
-    @saved = @todo.save
+    begin
+      @saved = @todo.save!
+    rescue ActiveRecord::RecordInvalid => exception
+      record = exception.record
+      if record.is_a?(Dependency)
+        record.errors.each { |key,value| @todo.errors[key] << value }
+      end
+      @saved = false
+    end
+
 
     # this is set after save and cleared after reload, so save it here
     @removed_predecessors = @todo.removed_predecessors
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -85,7 +85,7 @@ class TodosController < ApplicationController
       create_multiple
     else
       p = Todos::TodoCreateParamsHelper.new(params, current_user)
-      p.parse_dates() unless mobile?
+      p.parse_dates unless mobile?
       tag_list = p.tag_list
 
       @todo = current_user.todos.build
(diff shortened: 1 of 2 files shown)


C-Otto:
On my system running the tests will consume all of my RAM. This might be the reason for the Travis build to fail. I will investigate tomorrow.

lrbalt:
Could you rebase and see if it will pass on travis then?

C-Otto:
I will do, later this evening.

app/controllers/todos_controller.rb
```diff
@@ -423,7 +423,13 @@ def update
     update_dependencies
     update_attributes_of_todo
 
-    @saved = @todo.save
+    begin
+      @saved = @todo.save!
+    rescue ActiveRecord::RecordInvalid => exception
+      @saved = false
+      exception.record.errors.each { |key| exception.record.errors[key].each { |value| @todo.errors[key] << value } }
```
C-Otto: This line causes the value 'can't be blank' to be seen infinitely often when changing the description of an action to "". Could you explain how error messages are added to `@todo` and how this line should look like? I need to transport the error messages introduced in `dependency.rb:12` to `@todo` here, so that it is shown in the GUI.

C-Otto: `exception.record` and `@todo` are the same. This explains the nontermination.

C-Otto:
Now everything works, but I don't like the code. Could you please have a look into this?

I'm especially interested in how validation/error handling works. It seems validation is aborted if an entry is added to the `errors` attribute of the updated object. How can I write into the action's `errors` attribute when the verification inside `dependency.rb` fails? It seems the `successor` in `dependency.rb` object is neither `@todo` nor `@original_item` in `todos_controller.rb:update`.

lrbalt:
Do we have a test for this? It would be interesting to see if `exception.record == @todo`

C-Otto:
Yes, editing an action to have an empty description triggers exactly this (which is why my previous code ran into nontermination, copying error messages from the action to itself).

C-Otto:
Could you have a look into this and merge, please?

dnrce:
@C-Otto, I'd like to merge this but it should go into the 2.3_branch. Can you please rebase? After it's in 2.3 I can merge it down to master. Thanks!

dnrce:
Reopened #1754 but closing this in favor of a new PR targeting 2.3.

PR #1814, opened 2015-04-12
retain tags from action when transforming into project

### LINKED ISSUE #1448: Created project for "Make project" should have default tags set
*Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/1448*

When I decide to transform a todo into a projet using "Make project", the default context is set to the context of the todo. However, no default tags are set. The default tags should be taken from the todo, just as in the case of the context.

*Originally reported by carsten.otto on October 10, 2013 at 17:52:31 (+0000) against version 2.2.2*

C-Otto:
When creating a project out of an action, add the tags from the action as default tags for the project.

This fixes #1448.

### DIFF
--- a/lib/project_from_todo.rb
+++ b/lib/project_from_todo.rb
@@ -21,6 +21,7 @@ class ProjectFromTodo
       p.name = todo.description
       p.description = todo.notes
       p.default_context = todo.context
+      p.default_tags = todo.tag_list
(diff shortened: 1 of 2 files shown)


lrbalt:
merged in master. I'd like to keep the old behavior for 2.3

PR #1815, opened 2015-04-12
Update cucumber version

C-Otto:
Use newer version to fix problems with Firefox 35+.
`selenium-webdriver` older than 2.45 does not work with Firefox 35 and newer.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ group :test do
 
   gem "aruba", ">=0.5.4", :require => false
 
-  gem "selenium-webdriver"
+  gem "selenium-webdriver", '>=2.45'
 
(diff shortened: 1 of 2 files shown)


dnrce:
FYI, gem version updates generally don't need to involve Gemfile changes. Our standard is not to version things in the Gemfile as much as possible in `master`, and let Gemfile.lock be authoritative (otherwise it's kind of being redundant). Then when we create a release branch, we lock things down.

In this case, running `bundle update selenium-webdriver` and committing the changes to Gemfile.lock would have been sufficient.

C-Otto:
OK, thank you.

dnrce:
No, thank _you_. :smile:

PR #1820, opened 2015-04-13
Update project name and context name in 'new action' form

### LINKED ISSUE #1789: GUI error: adding action after renaming project/context
When I add an action in the project view after updating the project's name (but before reloading the page), the action is added, but an error message is shown.

### LINKED ISSUE #1796: GUI enhancement: renaming context in context view should also update context in form
Similar to the situation with projects, updating the name should also update the context used to add new actions.

C-Otto:
When adding a new action in the project view, after changing the project's name we need to set a hidden field in the form so that the updated project is used for new actions.

The same holds for the context view, where we need to update another hidden field after changing a context's name.

This fixes #1789.

In addition, this fixes #1796 by updating the context in the visible part of the form after changing the context's name in the context view.

### DIFF
--- a/app/assets/javascripts/tracks.js.erb
+++ b/app/assets/javascripts/tracks.js.erb
@@ -13,9 +13,6 @@ var TracksForm = {
         }
         toggleLink.parent().toggleClass('hide_form');
     },
-    set_project_name: function (name) {
-        $('input#todo_project_name').val(name);
-    },
     set_project_name_for_multi_add: function (name) {
         $('#multi_todo_project_name').val(name);
     },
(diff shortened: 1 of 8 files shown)


C-Otto:
@lrbalt The name `replace_context_form_with_updated_context` is misleading, and I fail to understand what it actually does. The comments tell me JQuery's `replaceWith` might be useful, though. http://api.jquery.com/replacewith/

C-Otto:
Could you have a look into this and merge, please?

dnrce:
This affects 2.2.3 as well as 2.3.rc1. @C-Otto, think you could rebase this to target 2.3_branch?

C-Otto:
Will do.

C-Otto:
See #1885

PR #1821, opened 2015-04-14
do not access nil

C-Otto:
This fixes a bug introduced recently in d08ffb3806081f849938d7c388855d3496b0e4fb.


C-Otto:
The bug is reverted.

PR #1823, opened 2015-04-14
Sort containers

### LINKED ISSUE #256: next actions with sensitive due-date doesn't automatically sorted with Ajax
*Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/256*

When I add new next action with due date set it goes to the bottom of the specified context unconditionally.

It's only gets sorted to a proper place of the list when I refresh the page.

*Originally reported by Anonymous on April 4, 2006 at 21:01:30 (+0800)*

C-Otto:
Sort containers in JavaScript, for example when updating an action or creating a new one.  This fixes #256.

### DIFF
--- a/app/assets/javascripts/tracks_pages.js
+++ b/app/assets/javascripts/tracks_pages.js
@@ -193,5 +193,21 @@ var TracksPages = {
 
     /* fade flashes and alerts in automatically */
     $(".alert").fadeOut(8000);
+  }, sort_container: function(container) {
+    function comparator(a, b) {
+        var contentA = $(a).attr('data-sort') || '';
+        var contentB = $(b).attr('data-sort') || '';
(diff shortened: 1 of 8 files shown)


C-Otto:
This only works after re-rendering the todo partials. Any idea how to enforce this once?

C-Otto:
This does not sort actions when they get unblocked (pending to active). I will work on it.

C-Otto:
I found some more places where we need to sort, this should cover about everything (feel free to tell me if I missed anything).

The JavaScript code has a few problems, and I might be looking into that soon. It seems we often try to change the contents of some action's `div` using `.html()`, which effectively adds a `div` inside the new `div` (both sharing the same id). Using `replaceWith()` might be the way to go, instead.

PR #1824, opened 2015-04-15
Convert a .should expectation to expect()

dnrce:
We're using RSpec 3 expectations syntax, in which `.should` is deprecated. This was throwing a deprecation warning.

FYI @C-Otto -- this was introduced in #1805.


C-Otto:
Thank you, I'll try to remember this.

PR #1828, opened 2015-05-05
Test with PostgreSQL

dnrce:
Addresses #1892


dnrce:
PostgreSQL needs some fixes, apparently. Closing until addressed.

PR #1864, opened 2015-05-13
replaceWith instead of html

### LINKED ISSUE #1841: bug: use replaceWith() instead of html()
In `app/views/todos/add_predecessor.js.erb` function `replace_updated_predecessor` the predecessor todo is updated. Here, the new content is written using `html()`, which nests the data provided in the argument inside the old content. As the new content also contains the todo's ID, I suspect that `replaceWith()` should be used here.

C-Otto:
Fix #1841


dnrce:
Rerunning failed Travis build...

PR #1865, opened 2015-05-13
Show links to context/project also for completed actions

### LINKED ISSUE #1840: GUI enhancement: add links to "completed actions" list
The "completed actions" list shows the project and context of each completed action (depending on the current view), but this information is not clickable. Adding links, similar to the list of active/pending actions with square brackets?, would be useful.

C-Otto:
Fix #1840

### DIFF
--- a/app/helpers/todos_helper.rb
+++ b/app/helpers/todos_helper.rb
@@ -297,17 +297,16 @@ module TodosHelper
   end
 
   def project_and_context_links(todo, parent_container_type, opts = {})
-    str = ''
+    links = ''
     if todo.completed?
-      links = []
-      links << todo.context.name unless opts[:suppress_context]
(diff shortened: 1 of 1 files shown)


dnrce:
Does this change mean the context name is no longer displayed in full for completed actions? I think that would be a loss of functionality.

C-Otto:
If the verbose option is not set, you just see [P] and [C]. You'd need to hover over it (or click) to see the project name/context name. If the verbose option is set, you see it.

Personally, I'm very interested in this information, and with the verbose option set I'm fine with the solution.

C-Otto:
This is how it looks like now: http://imgur.com/olA9yHJ,eAp4uYC#0

mattr-:
I was missing this just the other day! Thanks! :heart:

PR #1869, opened 2015-05-14
(2.4) create projects as reviewed at creation time

### LINKED ISSUE #1788: New projects should be marked as reviewed
While reviewing an existing project, I converted an (old) action into a project. This project shows in the review list, although it was just created. I think it should not need to be reviewed for a while (just as if it was reviewed just when it was created).

C-Otto:
Fixes #1788


mattr-:
:tada: :+1: for this!

mattr-:
I'm generally not a fan of ActiveRecord callbacks for something like this. Why not just override `save` instead?

C-Otto:
Is this a good idea? I just found this: http://stackoverflow.com/questions/13238616/rails-override-save-to-perform-selective-update

C-Otto:
Also: http://stackoverflow.com/questions/13469823/safe-to-override-save-method-in-activerecord

PR #1870, opened 2015-05-14
(2.4) in tag view, do not use more-or-less random context for the form

### LINKED ISSUE #1834: GUI enhancement: make choice of default context in starred view clear
I fail to understand how the default context is set in the starred view. I expect the topmost context (as defined in the context view) to be used, as it is done on the front page. However, it seems to be related to which contexts actually contain starred actions - and the choice of which context to use in the form is made without regarding the contexts' order.

C-Otto:
In the tag view (including starred) use the first (active) context in the form, based on the order of contexts as defined by the user. Picking the context of the oldest tagged (starred) action is not meaningful.

Fix #1834

### DIFF
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -635,9 +635,6 @@ class TodosController < ApplicationController
 
     # Set defaults for new_action
     @initial_tags = @tag_name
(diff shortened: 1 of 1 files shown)


C-Otto:
Also applies to 2.3.

PR #1873, opened 2015-05-14
(2.3) in tag view, do not use more-or-less random context for the form

### LINKED ISSUE #1834: GUI enhancement: make choice of default context in starred view clear

C-Otto:
See #1834 for a discussion
Fixes #1834


dnrce:
Can't take a behavior change in RC stage so marking this for 2.3.1.

PR #1874, opened 2015-05-14
(2.3) remove all tab indices

### LINKED ISSUE #1835: GUI enhancement: when adding a new note, tabulator should focus "add note" button
When I press tab in the new note text box, the "description" field of the "new action" form is selected. Instead, I expect the "add note" button below the note form to be selected.

C-Otto:
Fix #1835

### DIFF
--- a/app/helpers/todos_helper.rb
+++ b/app/helpers/todos_helper.rb
@@ -409,18 +409,6 @@ module TodosHelper
     return html
   end
 
-  def reset_tab_index
-    $tracks_tab_index = 0
-  end
-
-  def next_tab_index
-    # make sure it exists if reset was not called. Set to 20 to avoid clashes with existing form in sidebar
-    $tracks_tab_index ||= 20
-
-    $tracks_tab_index = $tracks_tab_index + 1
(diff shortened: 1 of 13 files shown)


dnrce:
Did you try simply removing the tab indexes? In the past that has been a good clean fix -- the browser is generally quite good about handling things, but Tracks has historically overspecified these.

C-Otto:
No, I will try. But I suggest to merge this nevertheless, as it solves an existing problem - there are more urgent things to clean in the Tracks code.

C-Otto:
I manually tested most forms and pages, and without any defined tab order it seems to be just fine. I guess it would be better to do some more manual testing, though.

I use Chrome.

mattr-:
Happy to see manual tab indexes go.

C-Otto:
Sorry, I did not mention it (it was buried in my own notes):

It seems the tab order is wrong when several action (10?) exist. Then, when editing an action, the TAB key also jumps to the panel on the right side of the page.

I'll try to investigate this, any ideas how this could be fixed without manual tab indexes?

Issue #1804, opened 2015-04-06
GUI enhancement: move action after changing context in project view


C-Otto:
Currently, the action is left shown in the old context's view. I expect it to appear in the new context view, and disappear from the old context view.

dnrce:
Fixed by #1875

PR #1875, opened 2015-05-14
(2.3) Provide data needed for partial views

### LINKED ISSUE #1836: GUI error: error message shown when editing action to new context
In the project view, when editing an action so that it belongs to a previously non-existing context, an error message "There was an error retrieving from server: error" is shown and the edit popup does not close. However, the new context is created and the action is updated correctly.

This does NOT happen in the starred view.

```
ActionView::Template::Error (undefined method `id' for nil:NilClass):
    2: @not_done = @not_done_todos.select {|t| t.context_id == context.id }
    3: # invalidate the cache every day because of staleness or
    4: # rendering of "due in x days" that change without touching updated at of the todo
    5: cache [context, source_view_key, current_user.date.strftime("%Y%m%d"), @tag_name] do
    6: -%>
    7:   <%=
    8:     render :partial => 'todos/collection',
  app/helpers/application_helper.rb:236:in `eval'
  app/helpers/application_helper.rb:236:in `eval'
  app/helpers/application_helper.rb:236:in `source_view_key'
  app/views/contexts/_context.html.erb:5:in `_app_views_contexts__context_html_erb__2233859530331264606_62088620'
  app/views/todos/_update_successful.js.erb:90:in `_app_views_todos__update_successful_js_erb___2350987662880956348_49463840'
  app/views/todos/update.js.erb:3:in `_app_views_todos_update_js_erb___2064830107471052161_49361520'
```

C-Otto:
We need to have `@project`/`@context` for the corresponding partial views. This was not provided when editing an existing action.

Furthermore, the JavaScript now also removes and re-adds an action when the context was changed, so that it appears in the correct container.

Fixes #1836 and #1804

### DIFF
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -425,6 +425,8 @@ class TodosController < ApplicationController
 
     @saved = @todo.save
 
+    provide_project_or_context_for_view
+
     # this is set after save and cleared after reload, so save it here
     @removed_predecessors = @todo.removed_predecessors
 
@@ -454,6 +456,15 @@ class TodosController < ApplicationController
           render :action => "edit", :format => :m
         end
       end
+    end
+  end
+
+  def provide_project_or_context_for_view
+    # see application_helper:source_view_key, used in shown partials
+    if source_view_is :project
+      @project = @todo.project
+    elsif source_view_is :context
+      @context = @todo.context
     end
   end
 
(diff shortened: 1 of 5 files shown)


dnrce:
@C-Otto, could you please add a test here?

C-Otto:
Test added.

dnrce:
Thanks!

PR #1876, opened 2015-05-14
(2.3) remove context if name is not set in update form

### LINKED ISSUE #1839: Bug: default context of project cannot be unset
When editing a project by removing its default context, the default context still remains set. I expect it to be unset (and not shown again when editing the project, and not be shown in the form used to create new actions).

C-Otto:
Fix #1839

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -337,6 +337,8 @@ class ProjectsController < ApplicationController
     if default_context_name.present?
       default_context = current_user.contexts.where(:name => default_context_name).first_or_create
       p['default_context_id'] = default_context.id
+    else
+      p['default_context_id'] = nil
     end
   end
 


app/controllers/projects_controller.rb
```diff
@@ -337,6 +337,8 @@ def default_context_filter
     if default_context_name.present?
       default_context = current_user.contexts.where(:name => default_context_name).first_or_create
       p['default_context_id'] = default_context.id
+    else
+      p['default_context_id'] = NIL
```
mattr-: using `NIL` here works, but I'd rather this would be lowercase `nil` so that people don't go looking for a constant named `NIL` in the code base.

C-Otto:
I updated the PR to use `nil` instead of `NIL`.

mattr-:
Cool. Thanks!

PR #1884, opened 2015-05-22
(2.3) Circular dependencies

### LINKED ISSUE #1754: Cyclic dependency

C-Otto:
Fix two issues already mentioned in pull request #1777. This should be added, as #1777 is imcomplete/buggy without. Also relevant for release of 2.3.

1) Drag-n-drop removes the dragged todo if adding the dependency fails.
2) Adding a dependency does not show any error message in the edit form.

Fixes #1754.

### DIFF
--- a/app/assets/javascripts/tracks.js.erb
+++ b/app/assets/javascripts/tracks.js.erb
@@ -301,7 +301,6 @@ var TodoItems = {
         /* Drag & Drop for successor/predecessor */
         var dragged_todo = ui.draggable[0].id.split('_')[2];
         var dropped_todo = this.id.split('_')[2];
-        ui.draggable.remove();
         $('.drop_target').hide(); // IE8 doesn't call stop() in this situation
 
         ajax_options = default_ajax_options_for_scripts('POST', relative_to_root('todos/add_predecessor'), $(this));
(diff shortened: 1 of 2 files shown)


C-Otto:
Milestone should be 2.3.1

C-Otto:
After dropping the action, an animation is shown where it "flies" back to its original position until the server sends the "added dependency" message. This is noticable especially for longer distances (on the screen between the two actions) and slow servers/slow network.

I don't really mind, but I feel I should mention this regression.

PR #1886, opened 2015-05-22
Ignore time, so that one can always defer to the due date

### LINKED ISSUE #1878: Defer action not possible: "after due date"
I have an action that is due tomorrow, and is shown today. I'd like to defer it to tomorrow, as I can't do anything about that today. However, I get the message "Defer date is after due date". This is wrong, and deferring should work.

(on current 2.4 master)

C-Otto:
Fix #1878


C-Otto:
Maybe we should drop the time when saving the due date (same for show_from?)?

mattr-:
Drop the time in the database you mean? I could potentially see use cases where the time would be useful, but I'm fine with dropping the time.

mattr-:
Also, :shipit: for this one, and perhaps another PR for dropping the time?

C-Otto:
Let's start with an issue, you can do the PR then :)

PR #1887, opened 2015-05-22
Activate successors on predecessor destroy

### LINKED ISSUE #1881: Convert action to project does not remove pending state of successor
When I create a project out of an action A which is the only predecessor of a pending action B, the dependency does not exist anymore. However, B's state still is pending.

C-Otto:
Move the activation code to the model, so that it also is used when converting an action into a project.

Fixes #1881


mattr-:
Seems good to me :+1:

mattr-:
@C-Otto could we get a rebase please?

C-Otto:
Rebased to 2.3.

PR #1901, opened 2015-07-29
Use box-sizing: border-box to style mobile input

matjack1:
Hi @mattr-,

I've tried your PR and I like it! I've just noticed that adding this commit you can better style the input fields in order to have a consistent left/right margin.

We could better organise the date select with flexbox, but I don't know if it's a problem to use it, even if only on mobile (http://caniuse.com/#feat=flexbox).


mattr-:
This is good stuff. I thought about flexbox briefly, but it's not well supported enough IMHO to put it into use here.

matjack1:
I agree @mattr-, but if you consider just the mobile version it might be good enough. All the mobile browsers are quite good with that and the worst that could happen is falling back on what we have now. If you want I can open a PR for you to check that.

PR #1903, opened 2015-08-04
Fix failing tests for non-UTC timezones

lrbalt:
There were several problems:
* Time.now returns the systems time, not the users time
* fixtures do not translate dates from timezone to utc, but stores the
  date verbatim
* calling a controller will set the timezone to the preference of the
  current_user. So it could be changed while you do not realize this. I
  fixed the failing test, but problems could be elsewhere

### DIFF
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -1071,10 +1071,10 @@ end
       if recurring_todo.todos.active.count == 0
 
         # check for next todo either from the due date or the show_from date
-        date_to_check = todo.due.nil? ? todo.show_from : todo.due
+        date_to_check = todo.due || todo.show_from
 
         # if both due and show_from are nil, check for a next todo from now
-        date_to_check = Time.zone.now if date_to_check.nil?
+        date_to_check ||= Time.zone.now
 
         if recurring_todo.active? && recurring_todo.continues_recurring?(date_to_check)
 
@@ -1087,7 +1087,7 @@ end
           # for tomorrow.
           date = date_to_check.at_midnight >= Time.zone.now.at_midnight ? date_to_check : Time.zone.now-1.day
 
-          new_recurring_todo = TodoFromRecurringTodo.new(current_user, recurring_todo).create(date.at_midnight)
+          new_recurring_todo = TodoFromRecurringTodo.new(current_user, recurring_todo).create(date)
         end
       end
     end
--- a/app/helpers/stats_helper.rb
+++ b/app/helpers/stats_helper.rb
@@ -5,7 +5,7 @@ module StatsHelper
   end
 
   def month_and_year_label(i)
-    t('date.month_names')[ (Time.now.mon - i -1 ) % 12 + 1 ]+ " " + (Time.now - i.months).year.to_s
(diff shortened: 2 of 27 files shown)


mattr-:
Looks good to me. @dnrce wanna give it your :eyes: before merge?

dnrce:
Great to be moving to zoned time objects and away from so much `.utc`!

I would recommend `Time.current` over `Time.zone.now` as the best general-use time helper to work with, though.

Unfortunately I have *soooo* little confidence in our time handling and time-based tests that I don't feel able to evaluate the "correctness" of this change any more concretely. However, the fact that all Travis tests are failing is a bad sign.

As to general code review:
* "This fixes failing tests when the timezone is" isn't a good commit summary -- perhaps "Fix failing tests for non-UTC timezones"?
* Fixup commits ("fix two regressions"?) should be squashed
* Ideally, whitespace changes and code cleanup should be separate commits

Are there related tickets? If so, they should also be referenced.

lrbalt:
never commit a timezone fix right after midnight :-) The tests were running before 01:00 CEST causing new corner cases.

Yes, timezone handling is not perfect. We set the servers timezone to the users preference for every request in the ApplicationController. This works fine, but not for tests where the tests and the app are running in the same process. Then stuff gets mixed up. 

I think we should not set the timezone in the ApplicationController, but use the time helpers when we need to present or parse a time for the user.

lrbalt:
As for you review comments:
* Time.zone.now vs. Time.current is a matter of taste I think. I like the first since it shows we took timezone into account, while it is difficult to remember if Time.now or Time.current uses timezones.
* I think I fixed the failure. It wasn't showing last night and this morning, but I think it was because of the time the tests ran on Travis (between 00:00 and 01:00 CEST)
* I'll research squashing the three commits into one. I've never had to :-)
* The whitespace thing is more difficult. Atom does it by default when I save. Difficult to keep them apart.

lrbalt:
I think you can merge this PR. The Travis failure is one timeout of cucumber.

mattr-:
I looked a little closer at the test failures. Two of them are indeed timeout errors. The other two, I'm not sure about. Here they are:

```
Couldn't find Context with 'id'=1 [WHERE `contexts`.`user_id` = ?] (ActiveRecord::RecordNotFound)
./app/controllers/todos_controller.rb:1013:in `block (2 levels) in determine_remaining_in_container_count'
./lib/tracks/source_view.rb:17:in `context'
./app/controllers/todos_controller.rb:1012:in `block in determine_remaining_in_container_count'
./lib/tracks/source_view.rb:43:in `source_view'
./app/controllers/todos_controller.rb:975:in `determine_remaining_in_container_count'
./app/controllers/todos_controller.rb:448:in `update'
features/context_edit.feature:10:in `And I have logged in as "testuser" with password "secret"'
```

and

```
undefined method `hidden?' for nil:NilClass (NoMethodError)
./app/models/todo.rb:229:in `hidden?'
./app/controllers/todos_controller.rb:1268:in `determine_changes_by_this_update'
./app/controllers/todos_controller.rb:447:in `update'
features/tagging_todos.feature:10:in `And I have logged in as "testuser" with password "secret"'
```

Any ideas?

mattr-:
Since the issues are with only a single build, I'm not going to worry about them. :shipit:

lrbalt:
I think the errors you mentioned are caused by the timeouts before. So adding a new context failed causing the lookup of that context to fail too

PR #1904, opened 2015-08-04
use sudo:false to migrate to new travis infra

lrbalt:


mattr-:
:+1: :shipit:

mattr-:
It would seem that the timeout errors have nothing to do with these changes, so merging anyways.

PR #1905, opened 2015-08-05
Add attachments to Todo model using Paperclip

lrbalt:
This is a squashed PR for adding attachments to the Todo model. Currently only emails received by the MessageGateway are saved as attachments. As such, this is the start of implementing a much asked for feature to add attachements to todos.


lrbalt:
I've squashed last two commits to make a nice PR out of this.

mattr-:
Build only fails on one thing, so this is not a systemic issue. Merging.

PR #1906, opened 2015-08-07
Completed todos are now selectable.

jrusnack:
So far, the whole row with a todo was draggable, with handle grip_span.
This means for normal todos the whole action is draggable, but user
must drag it by clicking on grip_span.

Grip span has not been rendered for completed todos, so the whole row
could be dragged, preventing user from selecting any text in completed
todo.

This commit makes grip_span rendered also for completed tasks, allowing
them to be dragged by grip_span, but the rest of the text is selectable
like for normal todos.

### DIFF
--- a/app/helpers/todos_helper.rb
+++ b/app/helpers/todos_helper.rb
@@ -269,11 +269,9 @@ module TodosHelper
   end
 
   def grip_span(todo=@todo)
-    unless todo.completed?
-      image_tag('grip.png', :width => '7', :height => '16', :border => '0',
-        :title => t('todos.drag_action_title'),
-        :class => 'grip')
-    end
+    image_tag('grip.png', :width => '7', :height => '16', :border => '0',
+      :title => t('todos.drag_action_title'),
+      :class => 'grip')
   end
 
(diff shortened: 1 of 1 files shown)


mattr-:
I understand the reasoning for the change, but why not just make it so that complete todos aren't movable?

jrusnack:
Simple - this has been in Track for some time and as a first time contributor, I have no way to tell if dragging completed actions is a feature. So instead of risking removing feature, I went for the smallest fix that keeps things working.

C-Otto:
I'd like completed todos to be non-moveable.

lrbalt:
This is not the right solution. drag and drop is used for changing a context. This is not useful for completed todos. We should disable drag and drop completely for completed todos

mattr-:
Since this isn't the right approach, I'm closing this pull request. The best approach would be to disable drag and drop completely for completed todos. @jrusnack is that something you could work on?

jrusnack:
@mattr- probably not - I don`t have much time to get into Tracks development right now and this patch works for me well.

PR #1911, opened 2015-08-10
Default project should be empty

matjack1:
I think there's been a merge conflict in the last batch and this has been overridden. It's re-fixing a recent fix :)

### DIFF
--- a/app/views/todos/_edit_form.html.erb
+++ b/app/views/todos/_edit_form.html.erb
@@ -22,7 +22,7 @@
 
   <div class="project_input">
     <label for="<%= dom_id(@todo, 'project_name') %>"><%= t('common.project') %></label>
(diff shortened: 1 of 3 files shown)


C-Otto:
Thank you!

dnrce:
Thanks! Can you add a test here?

mattr-:
My fault. Sorry. A test would definitely be good. @matjack1 is that something you think you could do soon?

matjack1:
Yes @dnrce and @mattr- I've just repushed the commit with a new test. I searched for an existing step definition that would be suitable for my test but I couldn't find it. If you have any suggestions they are welcome.
The test correctly fails without the fix, hope it's good :)

matjack1:
I don't get why the functional tests were failing. I've repushed to try again. I've Ruby 2.2 on my machine but I can't replicate the bug that was showing up in Travis. Let's see.

matjack1:
Can you have a look at this again? I just repushed on Travis, what do you think?

mattr-:
Yeah, we'll take a look at this soon.

mattr-:
Looks good to me. Thanks @matjack1!

PR #1917, opened 2015-08-19
remove old hack to set content type to application/html

lrbalt:
This prevents attachments to be seen as something different than
html content. For example, message/rfc822 for email was being
replaced by application/html causing wrong application to be
started on the client.


lrbalt:
the failures are those random timeout errors. I think it is ok to merge

PR #1918, opened 2015-08-19
speed improvements from fasterer gem

lrbalt:
The fasterer gem (https://github.com/DamirSvrtan/fasterer) suggest some speed improvements for Tracks. I've 'fixed' most of them,but left some that require more extensive refactoring.


mattr-:
:dash: Awesome. Thanks!

PR #1920, opened 2015-08-21
Add in Bootstrap for UI bits

mattr-:
This PR allows us to start playing with various bootstrap bits in the application, by wrapping those elements that should have access to bootstrap styling in a `bootstrap` class.

Nothing's changed in the UI yet, but this gives us the capability to make changes to the UI and start using bootstrap in the process.


Gemfile
```diff
@@ -42,6 +42,8 @@ gem 'bcrypt', '~> 3.1.7'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 # gem 'jbuilder', '~> 1.2'
 
+gem "bootstrap-sass"
```
dnrce: Let's be explicit about version here given that it's our views that are dependent on it. Tests could easily miss regressions, because the DOM will not have changed. So we should have to be deliberate about bumping this.

app/assets/stylesheets/application.css.scss
```diff
@@ -14,3 +14,8 @@
  *= require_tree ../../../vendor/assets/stylesheets
  *= require tracks
  */
+
+.bootstrap {
```
dnrce: Is the use of `.bootstrap` intended to be permanent, or is this to enable a gradual rollout during development?

dnrce:
Woo!

Gemfile
mattr-: Doh. I thought I'd done that already. I'll update it.

app/assets/stylesheets/application.css.scss
mattr-: This is to enable a gradual rollout.

Gemfile
mattr-: And done. :dancers:

matjack1:
I like it :+1:

PR #1922, opened 2015-08-26
Add a configuration setting for serving static assets

### LINKED ISSUE #1896: Allow static assets to be served without editing environment files
Should not be served by default, but move the configuration into `site.yml`.

mattr-:
This new setting will allow us to control in `site.yml` whether or not our static assets will be served by Rails or by the app server that is proxying requests to Rails. This is set to `false` by default so that the proxying server is responsible for serving the assets.

Fixes #1896


config/site.yml.tmpl
```diff
@@ -25,6 +25,11 @@ secure_cookies: false
 #
 secret_token: "change-me"
 
+# By default, tell Rails to not serve static assets. The best practice is to
```
dnrce: I think the first line should describe the functionality of this option, regardless of how it's set. Then the true/false values, and the reason for the default, will be on the following lines.

config/environments/production.rb
```diff
@@ -20,7 +20,7 @@
   # config.action_dispatch.rack_cache = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this).
-  config.serve_static_assets = false
+  config.serve_static_assets = SITE_CONFIG['serve_static_assets']
```
dnrce: Should default to `false` if unset.

config/site.yml.tmpl
```diff
@@ -25,6 +25,11 @@ secure_cookies: false
...
+# have a proxying web server such as Apache or Nginx serve static assets
+# (images, stylesheets, javascript) for you. Change this to 'true' if you want
+# Rails to be responsible for serving the static assets
+serve_static_assets: false
```
dnrce: This can be commented out in the template.

dnrce:
Awesome.

Add doc updates here, too -- a changelog entry and the "configure variables" section of the installation guide. And possibly an upgrade note, but I'm not sure that's necessary.

dnrce:
Side note: I was puzzled for a minute as to why this triggered two Travis builds and then I realized it was because this is a branch in the main repo. My feeling is that standard workflow is to open PRs from your fork unless it's something long-running that others will contribute to. This gives parity across all contributors and keeps the main repo branches tidy. Just a thought.

dnrce:
The upgrade instructions definitely need a change. They currently mention editing production.rb.

config/environments/production.rb
mattr-: Good point. Hadn't thought about that case.

config/site.yml.tmpl
mattr-: Is it the bit that says 'By default' that makes this bad? The functionality is to have Rails serve or not serve static assets. I'm a bit confused as to how the first line doesn't describe the functionality of the option.

Definitely need the reason for the default though. I'll add that in.

config/site.yml.tmpl
mattr-: For new installs, wouldn't we always want this to be set to something vs. relying on a fallback? Or is it that commenting this out means that the user doesn't have to worry about changing it?

mattr-:
I don't currently have my own fork, which is why this is just a branch in the main repo. I'll create one though, update these changes, and then close this PR and make a new one when it's ready.

config/site.yml.tmpl
dnrce: Yeah, the message is good, it's just a style suggestion that the first line be only a summary of what the option _does_ rather than a description what it's _set to_. For example, "Serve static assets (images, stylesheets, scripts) in production," followed by an explanation of the defaults, etc., on subsequent lines.

config/site.yml.tmpl
dnrce: The latter. I believe settings should only be populated if they are overrides. If you look at the Rails environment files there are a number of options like this.

dnrce:
No need to close this, just a thought for the future.

mattr-:
Made some updates. Happy to change whatever else needs to be done. :smiley:

config/environments/production.rb
```diff
@@ -20,7 +20,7 @@
   # config.action_dispatch.rack_cache = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this).
-  config.serve_static_assets = false
+  config.serve_static_assets = SITE_CONFIG['serve_static_assets'] || false
```
dnrce: :thumbsup:

config/site.yml.tmpl
```diff
@@ -25,6 +25,11 @@ secure_cookies: false
 #
 secret_token: "change-me"
 
+# Configure how static assets (images, stylesheets, etc.) will be served.
+# The best practice is to have a proxying web server such as Apache or Nginx
+# serve static assets (images, stylesheets, javascript) for you. Change
+# this to 'true' if you want Rails to be responsible for serving the static assets
+# serve_static_assets: false
```
dnrce: :ok_hand: perfect

doc/upgrading.md
```diff
@@ -15,7 +15,7 @@ That said. To upgrade:
 7. Precompile your static assets (css, javascript, etc.) by running `bundle exec rake assets:precompile RAILS_ENV=production`.
 8. Run `bundle exec rails server -e production` inside your Tracks 2.4devel directory to start up Tracks. Or use `-e development` if you intend to try your changes and get more log info.
 
-Please note that if you intend to deploy Tracks using its included web server, you’ll need to change `config.serve_static_assets` to `true` in `config/environments/production.rb` in order for the images, stylesheets, and javascript files to be served correctly.
+Please note that if you intend to deploy Tracks using its included web server, you’ll need to copy over the `serve_static_assets` option from `config/site.yml.tmpl`,  uncomment the option, and set it to `true` in `config/site.yml` in order for the images, stylesheets, and javascript files to be served correctly.
```
dnrce: For 2.3->2.4 this ought to to instruct users to transfer the setting from production.rb to site.yml.

dnrce:
Sweet.

The installation guide also needs a change, and this can all be squashed.

mattr-:
Squashed all the multiple commits down to three and updated the installation guide.

mattr-:
Should I squash this down to a single commit instead?

doc/upgrading.md
```diff
@@ -15,7 +15,7 @@ That said. To upgrade:
 7. Precompile your static assets (css, javascript, etc.) by running `bundle exec rake assets:precompile RAILS_ENV=production`.
 8. Run `bundle exec rails server -e production` inside your Tracks 2.4devel directory to start up Tracks. Or use `-e development` if you intend to try your changes and get more log info.
 
-Please note that if you intend to deploy Tracks using its included web server, you’ll need to change `config.serve_static_assets` to `true` in `config/environments/production.rb` in order for the images, stylesheets, and javascript files to be served correctly.
+Please note that if you intend to deploy Tracks using its included web server, you’ll need to copy over the `serve_static_assets` option into your `config/site.yml` from `config/site.yml.tmpl`,  uncomment the option, and set it to the the value you had for `config.serve_static_assets in `config/production.rb` in order for the images, stylesheets, and javascript files to be served correctly.
```
dnrce: This is a bit unwieldy. Is it enough to say that the option should now be set in `config/site.yml` instead of `config/environments/production.rb`?

dnrce:
This is definitely subjective, but yeah, one commit would be my preference. To me this is all one logical change -- none of these parts would really stand on their own. If I ever did a `git blame` and looked at the commit I would want to see all of this.

If you squash things further, I think your first commit message describes everything nicely.

mattr-:
Squashed into all one commit, and updated the upgrading guide to be a bit less verbose.

dnrce:
:boom:

PR #1923, opened 2015-08-27
Port top navigation bar to Bootstrap

matjack1:
I've tried to keep this PR as small as possible to start the new Bootstrap UI project. The idea is to swap the navigation with this version that is responsive. Is not adding anything yet to pages that are not responsive, but now we can start to port one page after another.

A couple of comments regarding the code:
- I've introduced font-awesome inspired by the new-gui branch, but it's not essential. I think it's better looking and better to use than what we had before, but it's just my opinion.
- in the bootrstrap_and_overrides CSS I'm now just making bootstrap work with the current CSS. Once the transition is complete we could delete this file.
- I had to rename the badge class because it conflicted with a badge class in Bootstrap.
- I've not ported the colours from the old navigation, but I think it would be a good idea, to preserve a bit the style. I'm opening the PR to have a conversation on the code, but it's still a WIP.
- if you could try on your computers and mobiles would be great!

One thing that I think it's good on this new version is that it's a bit smaller than the previous nav, saving 20px at the top, which is nice.

PR #1924, opened 2015-08-28
Namespace the legacy CSS

dnrce:
This allows Bootstrap changes to be made in almost total isolation. The only element with overlapping/shared styles will be `body`.

### DIFF
--- a/app/assets/javascripts/tracks.js.erb
+++ b/app/assets/javascripts/tracks.js.erb
@@ -833,7 +833,8 @@ var RecurringTodosPage = {
           RecurringTodosPage.hide_all_recurring();
           RecurringTodosPage.reset_radio();
           $('#recurring_daily').show();
-        }
+        },
+        appendTo: "#content"
       });
 
       /* change recurring period radio input on new form */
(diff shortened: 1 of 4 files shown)


matjack1:
Thanks @dnrce, this is a nice idea! I'll rebase my branch over this one and simplify that a lot.

dnrce:
@mattr- what do you think? This seems like a companion to your #1920.

dnrce:
I made some changes to improve conflicts over `body`.

@matjack1, have another look here.

matjack1:
:+1: thanks!

mattr-:
I'm good with this. Should we close this one and just let #1923 1923 do all the work for us?

dnrce:
I wouldn't mind seeing this merged independently if it's ready to go. That way the other PR will be self-contained.

mattr-:
Does this need a changelog entry or anything like that? or do we want to save it for when we do more UI work and just not worry about such granular changes such as these?

dnrce:
This has no user-facing impact so I don't think it needs an entry. #1923 should probably have an entry.

matjack1:
I'd say to merge this PR first, so that #1923 it's smaller and easier to debug if we introduce problems.

mattr-:
:boom: Merged.

PR #1925, opened 2015-08-29
Increase the maximum password length

### LINKED ISSUE #1921: Increase the maximum password length
The passwords are limited to 40 characters.
I don't like the limit, so is it possible to increase it?

SuperNascher:
This resolves #1921

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -100,7 +100,7 @@ class User < ActiveRecord::Base
 
   validates_presence_of :login
   validates_presence_of :password, if: :password_required?
-  validates_length_of :password, within: 5..40, if: :password_required?
+  validates_length_of :password, within: 5..128, if: :password_required?
   validates_presence_of :password_confirmation, if: :password_required?
   validates_confirmation_of :password
   validates_length_of :login, within: 3..80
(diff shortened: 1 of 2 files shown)


dnrce:
lenght

dnrce:
I know there is at least one test that will need to be updated.

dnrce:
Also, can you please adjust the commit message to make it clear that this only the _maximum_ length being adjusted? "Increase the maximum password length", for example.

SuperNascher:
Ok

dnrce:
It looks like `test_validate_long_password` is the only failing test. Please update https://github.com/TracksApp/tracks/blob/9a18705f70eafc6ac442713203e8b25d7285d61d/test/models/user_test.rb#L47-L48.

dnrce:
Actually, it probably makes sense to limit the length to 72 characters. Bcrypt ignores anything after 72. See Devise: https://github.com/plataformatec/devise/commit/c7cc91cb65687e0325d817472a9da9b0c3894b41

SuperNascher:
I have set the maximum length to 72 characters on both files.

dnrce:
:thumbsup: Awesome. Thanks!!

PR #1939, opened 2015-11-18
Bump mysql2 to 0.3.20

mattr-:
This allows us to support MySQL 5.7.

Keep it locked at the 0.3.x version due to problems with other apps and
the 0.4.x versions.


mattr-:
No comments after nine days?!? Merging. :smiley:

PR #1942, opened 2015-11-27
Bump mysql2 to 0.3.20

mattr-:
I recently bumped my local MySQL install to 5.7, which didn't work with earlier versions of the gem. 

Keep it locked at the 0.3.x version due to problems with other apps and the 0.4.x versions.


mattr-:
doh! I already did this. :disappointed:

PR #1943, opened 2015-11-27
Redesign the login page

mattr-:
Makes use of both bootstrap and a rendering of the Tracks logo to provide a bit better experience in term of brand identity as well as also supporting mobile better.

There are a few things left to do here before this gets merged, but I wanted to provide something for feedback, since this is a large departure from our previous login page:

 - [x] Fix the flash rendering to use bootstrap styles and be colored (it's currently unstyled)
 - [x] Remove the use of flexbox for better browser support (Firefox and IE8+)
 - [x] Verify the cucumber scenarios aren't broken. Fix them if they are.
 - [x] Double check rendering on IE 8+
 - [x] Fix Safari 7 rendering issue
 - [x] Check IE 10 rendering
 - [x] Check IE 11 rendering

### Desktop (Chrome on OS X)
Desktop - Chrome on OS X

### Nexus 5 Emulation with Chrome
Nexus 5 Emulation with Chrome

### iPad Emulation with Chrome
iPad Emulation with Chrome

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -25,6 +25,8 @@
 .bootstrap {
   @import "bootstrap-sprockets";
   @import "bootstrap";
+  @import "tracks-logo";
+  @import "login";
 
   // resets normally applied to body
   // TODO: remove when dropping the .bootstrap namespace
--- a/app/assets/stylesheets/login.css.scss
+++ b/app/assets/stylesheets/login.css.scss
@@ -6,6 +6,7 @@ body {
   @include make-xs-column(12);
   @include make-sm-column(6);
   @include make-sm-column-offset(3);
+  padding: 10px;
 }
 
 .login-wrapper {
--- a/app/assets/stylesheets/login.css.scss
+++ b/app/assets/stylesheets/login.css.scss
@@ -1,150 +1,50 @@
-body { background-color: #fff; color: #333; }
-
-body, ol, ul, td {
-    font-family: verdana, arial, helvetica, sans-serif;
-    font-size: 12px;
-    line-height: 18px;
-    background: #eee;
+body {
+  background-color: rgba(0, 0, 0, 0.75);
 }
 
-p {
-    background: #eee;
+.login-flash {
+  @include make-xs-column(12);
+  @include make-sm-column(6);
+  @include make-sm-column-offset(3);
 }
 
-a, a:link, a:active, a:visited {
-    color: #cc3334;
-    text-decoration: none;
-    padding-left: 1px;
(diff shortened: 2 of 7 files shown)


mattr-:
I discovered that CSS gradients aren't supported in IE until version 10. IE8 and IE9 users account for approximately 2% of the global user base. Are we ok with dropping support for those browsers in version 2.4?

dnrce:
Sharp!

How does IE degrade?

Some thoughts:

My feeling is that Tracks should move away from the "10" logo, because I don't think it makes much sense, and toward the check mark instead.

I don't know whether this shadowed raised panel styling will be consistent with any UI elements in the actual app.

I had been thinking we might even want to do away with a separate login layout in the move to Bootstrap.

mattr-:
>How does IE degrade?

:laughing: Not well. I'll pull screenshots from the VM and post them tomorrow.

> My feeling is that Tracks should move away from the "10" logo, because I don't think it makes much sense, and toward the check mark instead.

I'm cool with that. I was mostly trying to reuse the existing branding rather than try to come up with something new.

> I had been thinking we might even want to do away with a separate login layout in the move to Bootstrap.

I will admit to thinking about that for a minute and then deciding that I would try to make the most minimal amount of change possible. I'm sure it can be done. I think it's a great idea. I just didn't do it as part of this PR.

mattr-:
Here's IE 8:
new login ie8

Here's IE 9:
new login ie9

dnrce:
Ha! Go, IE.

I guess I didn't answer your question directly before but I'm totally happy with dropping IE8 & 9 support. Users will still have Tracks 2.3 available to them.

mattr-:
:+1: Cool.

mattr-:
I intend to squash this into a better story commit wise once it's good to go, just putting the individual commits here so that it's easier to review as I make changes.

mattr-:
Also, IE 10 and IE 11 both look fine:

IE 10:
screen shot 2015-12-13 at 7 49 04 pm


IE 11:
screen shot 2015-12-13 at 7 52 11 pm

mattr-:
:boom: All done. I think. Here's a screenshot with the flash styled properly.

screen shot 2015-12-15 at 8 03 34 pm

mattr-:
I can't reproduce the cucumber test failures locally by either running the scenario that's failing by itself or the whole cucumber suite all together (well, at least what's completed so far). I think i'll wait for the full suite to finish on my machine and if it does so without errors, go ahead and merge this.

mattr-:
got an error with the tagging todos completely unrelated to anything this changes. Merging.

C-Otto:
Please have a look at issue #1954

PR #1946, opened 2015-12-05
when recurring todos always use midnight for "show from"

### LINKED ISSUE #1940: Deferred action with "Show Today"
I have a recurring todo which is supposed to be shown today. Indeed, I see it in the tickler/deferred view/... with a box containing "Show Today". Why isn't that shown right now?

The `show_from` column contains `2015-11-21 20:55:22.2137`, which is in a few hours. I guess the H:M:s part of recurring todos should be set to midnight (or treated as such?).

C-Otto:
fixes #1940

### DIFF
--- a/app/models/recurring_todos/abstract_recurrence_pattern.rb
+++ b/app/models/recurring_todos/abstract_recurrence_pattern.rb
@@ -141,7 +141,7 @@ module RecurringTodos
         get_due_date(previous) - show_from_delta.days
       when 'show_from_date'
         # Leave due date empty
-        get_next_date(previous)
+        get_next_date(previous).at_midnight
       end
     end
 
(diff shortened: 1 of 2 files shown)


mattr-:
Does this need to go to the 2.3 branch as well?

C-Otto:
I will investigate.

mattr-:
If this does need to go to the 2.3 branch, it'll have to be modified as the current patch doesn't apply cleanly. I think we'll just leave this as a 2.4 thing for now.

C-Otto:
Cherry-picking worked just fine, see #1948

C-Otto:
This only works for recurring todos with a "show from" setting, not a "due date". I'll fix.

PR #1948, opened 2015-12-10
when recurring todos always use midnight for "show from"

### LINKED ISSUE #1940: Deferred action with "Show Today"

C-Otto:
fixes #1940


C-Otto:
This only works for recurring todos with a "show from" setting, not a "due date". I'll fix.

C-Otto:
Fixed.

PR #1959, opened 2016-01-13
Update the footer to be more responsive

mattr-:
Matches the style of the existing footer and makes the links move to the next line instead of wrapping if the viewport is too small for all of the footer to fit on one line.

Also moves the footer over to bootstrap styles

This makes the footer work for both the login screen and other pages that aren't fully bootstrapped yet. It also solves a design issue that is really nitpicky but was driving me nuts. :smiley:

### DIFF
--- a/app/assets/stylesheets/legacy.css.scss
+++ b/app/assets/stylesheets/legacy.css.scss
@@ -402,15 +402,6 @@ input.item-checkbox {
     border-top: 1px solid #999;
 }
 
-#footer {
-    clear: both;
-    font-size: 85%;
-    text-align: center;
-    color: #999;
-    margin: 20px 310px 5px 20px;
-    padding: 0px;
-}
-
 a.footer_link {
     color: #cc3334;
     font-style: normal;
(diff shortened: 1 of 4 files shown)


mattr-:
Screenshots!! :tada: :confetti_ball: 

**Old Footer on the Login Page**
screen shot 2016-01-12 at 9 40 46 pm

**New Footer on the Login Page**
screen shot 2016-01-12 at 9 40 58 pm

mattr-:
And two more versions to show how it behaves when the wrapping is triggered:

**Tablet version**
screen shot 2016-01-13 at 8 20 46 am

**Phone version**
screen shot 2016-01-13 at 8 20 55 am

dnrce:
Sweet!

Can you remove the legacy `#footer` styles as well?

mattr-:
oh, yeah. will do.

mattr-:
Done! Should be :+1:

dnrce:
:ship: :it:

C-Otto:
This is what I see:

C-Otto:
Ah, sorry. I always forget to regenerate the assets. Now it's fine, thank you!

PR #1961, opened 2016-01-17
Sort current projects

### LINKED ISSUE #1960: Improvement: Sort current projects in review page by review-age
I stopped reviewing all my (77) projects in one go, and instead review a part of them on a daily basis. Now I'm at a state where all projects are current, but can't see which I have to review next.

The current list is ordered like on the right panel, which is not helpful. I'd love to see the projects, at least in the `current` list, ordered by `last review date`. This date should also be shown.

C-Otto:
I don't understand what the `.load` is supposed to do.

We could also discuss to also sort the `review`, `stalled` and `blocked` lists.

I'd prefer to not sort the `review` list, as
- all of the projects need to be reviewed now/soon-ish
- the user might have defined an order based in priority

This fixes #1960

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -58,7 +58,6 @@ class ProjectsController < ApplicationController
     @source_view = params['_source_view'] || 'review'
     @page_title = t('projects.list_reviews')
     @projects = current_user.projects.load
-    @contexts = current_user.contexts.load
     @projects_to_review = current_user.projects.select  {|p| p.needs_review?(current_user)}
     @stalled_projects = current_user.projects.select  {|p| p.stalled?}
     @blocked_projects = current_user.projects.select  {|p| p.blocked?}
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -58,9 +58,9 @@ class ProjectsController < ApplicationController
     @source_view = params['_source_view'] || 'review'
     @page_title = t('projects.list_reviews')
     projects = current_user.projects
-    @projects_to_review = current_user.projects.select  {|p| p.needs_review?(current_user)}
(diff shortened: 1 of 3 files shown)


app/controllers/projects_controller.rb
```diff
@@ -57,12 +57,11 @@ def index
...
-    @contexts = current_user.contexts.load
-    @projects_to_review = current_user.projects.select  {|p| p.needs_review?(current_user)}
-    @stalled_projects = current_user.projects.select  {|p| p.stalled?}
-    @blocked_projects = current_user.projects.select  {|p| p.blocked?}
-    @current_projects = current_user.projects.uncompleted.select  {|p| not(p.needs_review?(current_user))}
+    projects = current_user.projects
+    @projects_to_review = projects.select  {|p| p.needs_review?(current_user)}
+    @stalled_projects = projects.select  {|p| p.stalled?}
+    @blocked_projects = projects.select  {|p| p.blocked?}
+    @current_projects = current_user.projects.uncompleted.select { |p| not (p.needs_review?(current_user)) }.sort_by { |p| p.last_reviewed || Time.zone.at(0) }
```
mattr-: can we use `projects` here instead of calling `current_user.projects`?

mattr-:
One small comment, but PR looks good otherwise, and the tests all pass locally. Going to blame the CI failures on Travis for now.

app/controllers/projects_controller.rb
C-Otto: Yes, I'll fix. This was not possible in my previous version, where I sorted all projects (as the resulting list/array/... does not offer the "uncompleted" method).

C-Otto:
Fixed according to comment.

mattr-:
Awesome. Thanks! :heart:

PR #1967, opened 2016-01-23
add target="_blank" for generated links

### LINKED ISSUE #1747: Links in the notes don't open a new browser tab
Use case:
- I use Google Apps for Business, and regularly link to the tasks I have to take care of from within the notes.
- while links copied from the Mail tab are automatically recognized within tracks, they are not opened in a new tab
- so if I click on the link, it opens in the tab previously occupied by tracks.

C-Otto:
This causes the links to be opened in a new window/tab.
Existing (rendered) notes are not affected by this.

Fixes #1747

### DIFF
--- a/lib/tracks/utils.rb
+++ b/lib/tracks/utils.rb
@@ -25,14 +25,17 @@ module Tracks
     def self.render_text(text)
       rendered = auto_link_message(text)
       rendered = textile(rendered)
-      rendered = helpers.auto_link(rendered, :link => :urls)
+      rendered = helpers.auto_link(rendered, :link => :urls, :html => {:target => '_blank'})
 
-      # add onenote and message protocols
-      config = Sanitize::Config.merge(Sanitize::Config::RELAXED,
-        :protocols => { 'a' => {'href' => Sanitize::Config::RELAXED[:protocols]['a']['href'] + ['onenote', 'message']}}
-        )
+      relaxed_config = Sanitize::Config::RELAXED
+      config = relaxed_config
 
-      rendered = Sanitize.clean(rendered, config)
+      # add onenote and message protocols, allow a target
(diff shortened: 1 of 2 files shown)


lib/tracks/utils.rb
```diff
@@ -25,14 +25,20 @@ def self.auto_link_message(text)
...
       # add onenote and message protocols
-      config = Sanitize::Config.merge(Sanitize::Config::RELAXED,
-        :protocols => { 'a' => {'href' => Sanitize::Config::RELAXED[:protocols]['a']['href'] + ['onenote', 'message']}}
-        )
+      a_href_config = relaxed_config[:protocols]['a']['href'] + %w(onenote message)
+      config = Sanitize::Config.merge(config, :protocols => {'a' => {'href' => a_href_config}})
+
+      # allow a target
+      a_attributes = relaxed_config[:attributes]['a'] + ['target']
+      config = Sanitize::Config.merge(config, :attributes => {'a' => a_attributes})
```
mattr-: I think you should be able to do this with only one merge call.
```
config = Sanitize::Config.merge(config, :protocols => ...,
                                                               :attributes => ...)
```

C-Otto: Done, thanks.

mattr-:
Looking good now. Thanks!

dnrce:
I should have commented before but I'm opposed to this. I don't like forcing window management strategies on users. Browsers make it easy enough to open links in new tabs.

mattr-:
I like the idea that clicking a link in Tracks doesn't actually take me away from Tracks, but I'm willing to revert this if you think we should just let users manage this on their own.

C-Otto:
A configuration option would also be possible.

PR #1972, opened 2016-01-29
Make glyphicons available for use in the UI

mattr-:
Ideally, it would have been possible to wrap this in the bootstrap only section of the application stylesheet, but the icons won't work when glyphicons are limited to bootstrap only bits of UI, so import them globally.

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -49,3 +49,12 @@ body {
   color: $text-color;
   background-color: $body-bg;
 }
+
+@font-face{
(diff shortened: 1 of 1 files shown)


dnrce:
Any reason not to just import `_glyphicons.scss` in its entirety?

dnrce:
Also FWIW I think Font Awesome might be a better bet. I like that the Glyphicons are built into bootstrap but it's a more limited icon set.

mattr-:
I believe we could import `_glyphicons.scss` at the top level, although I like the idea of using Font Awesome better, so I'm going to go with that and close this pull request.

PR #1974, opened 2016-01-30
Update the preferences pages to Bootstrap

mattr-:
No behavior changes, and all the features passed locally.

- [x] Double check date and time preferences updating
- [x] Make sure that all the other preferences work

PR #1975, opened 2016-01-30
Use jquery-ui-rails gem instead of vendored jQuery UI

dnrce:
I opted to include only individual modules and I _think_ I covered everything in use. I'm running the Cucumber tests right now but they probably don't provide ideal coverage.

The only visual change should be that this abandons Tracks' custom jQuery UI theme, defaulting to gray. This suits me, as I find the yellow incongruous, but I'm curious how others feel. If you see other changes, please flag them.


dnrce:
No need for `tabs` thanks to #1974!

mattr-:
I'm good with this, and I like the gray. :+1:

PR #1977, opened 2016-01-31
Add pry for debugging

dnrce:
I add this manually too often, so let's get it in the Gemfile.

Rails recommends `byebug` but it's Ruby-2.2-only and I still often find myself working on 1.9.3. We can switch when we go to Rails 5.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,10 @@ group :development do
   gem "activerecord-deprecated_finders"
 end
 
+group :development, :test do
+  gem 'pry'
+end
+
 group :test do
   gem "factory_girl_rails"
   gem "capybara"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -59,6 +59,10 @@ group :development do
   gem "activerecord-deprecated_finders"
 end
 
+group :development, :test do
+  gem 'pry'
```
mattr-: Could we add a `require: false` here? This means that anybody wanting to use pry would have to do `require 'pry'; binding.pry` instead of just `binding.pry` but for the default case of not doing any debugging, I like the lack of additional footprint when running the server or tests.

dnrce: What footprint do you mean? Load time should be a non-issue with spring.

mattr-: Oh, right. Sorry. This is the only project I've worked on that actually uses Spring.

PR #1978, opened 2016-01-31
Add RuboCop

dnrce:
* Exclude bin/ because it's generated code
* Max line length of 120 characters
* Documentation not required
* Prefer double-quoted strings

No real enforcement or code modification at this time, but I'd like to get this in place as a helper for future work.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,15 @@
+Rails:
+  Enabled: true
+
+AllCops:
+  Exclude:
+    - bin/*
+
+Metrics/LineLength:
+  Max: 120
+
+Style/Documentation:
+  Enabled: false
+
+Style/StringLiterals:
+  EnforcedStyle: double_quotes
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ group :development do
   gem "rack-mini-profiler"
   gem "rack-dev-mark"
   gem "activerecord-deprecated_finders"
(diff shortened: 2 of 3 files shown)


dnrce:
OK, @mattr-, this PR currently specifies a preference for double quotes, but here are my arguments for single quotes instead:
* *Don't have to use the shift key all the time to type them*
* Less visual clutter
* Conveys intent when using interpolated strings (I don't care about this point nearly as much)

C-Otto:
Just to add to the discussion: IntelliJ's default is single quotes unless double quotes are necessary.

mattr-:
All good points. My preference for double quotes comes from three places:

 - I've never had single quotes mean a string in other languages I worked in before Ruby, so part of it is habit.
 - Using double quotes reduces the change necessary when needing to add interpolation to an existing string.
 - Double quotes everywhere is more consistent when dealing with interpolated vs. non-interpolated strings. There's one less thing to remember when working in the codebase.

That said, I don't hold these opinions very strongly, so if the preference is for single quotes, then I'll go with that. :grinning:

C-Otto:
I lack experience and I'm fine with both. IntelliJ can be reconfigured :)

mattr-:
And because I merged the PR with Pry in it, this now needs a rebase.

mattr-:
Merged this manually. ddffe8f...80ec0b6

PR #1981, opened 2016-01-31
Remove project hidden state

### LINKED ISSUE #1082: get consistent hidden states
*Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/1082*

todo.hidden? looks at todo.state, context.hidden? and project.hidden?
moving a todo to a hidden project sets the state of the todo to project_hidden
moving a todo to a hdden context does not change the state

I'd propose to use only one hidden state for todo that is set when a todo is added/moved to a hidden project or a hidden context. This cleans up a lot of logic trying to figure out if the todo is hidden or not (including named scopes of todo)

*Originally reported by lrbalt on January 1, 2011 at 22:42:55 (+0500)*

### LINKED ISSUE #1423: Pending actions not shown for hidden projects
*Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/1423*

If action A depends on action B, both in project X, then action A should be shown in the "Deferred/pending actions in this project" part of project X. This is the case if the project is marked as active. However, for hidden projects both actions are shown in "Actions in this project".

*Originally reported by carsten.otto on May 5, 2013 at 12:41:31 (-0500) against version 2.2.2.git-master*

C-Otto:
I intend to clean this up a bit, but right now I'd love to get some feedback.

Fixes #1423
Fixes #1082

### DIFF
--- a/db/migrate/20160131233303_remove_project_hidden_state_from_todos.rb
+++ b/db/migrate/20160131233303_remove_project_hidden_state_from_todos.rb
@@ -1,0 +1,42 @@
+class RemoveProjectHiddenStateFromTodos < ActiveRecord::Migration
+  class Todo < ActiveRecord::Base
+    has_many :successor_dependencies, :foreign_key => 'successor_id', :class_name => 'Dependency'
+    belongs_to :project
+  end
+
+  class Project < ActiveRecord::Base
+  end
+
+  class Dependency < ActiveRecord::Base
+    belongs_to :predecessor, :foreign_key => 'predecessor_id', :class_name => 'Todo'
+  end
+
+  def self.up
+    Todo.where(state: 'project_hidden').find_each do |todo|
+      if todo.show_from.present?
+        todo.state = 'deferred'
+      elsif has_uncompleted_predecessor(todo)
+        todo.state = 'pending'
+      else
+        todo.state = 'active'
+      end
+      todo.save
+    end
+  end
+
+  def self.down
+    Project.where(state: 'hidden').find_each do |project|
+      Todo.where(project_id:  project.id).where.not(state: 'completed').find_each do |todo|
+        todo.state = 'project_hidden'
+        todo.save
+      end
+    end
+  end
+
+  def has_uncompleted_predecessor(todo)
+    todo.successor_dependencies.each do |dependency|
+      return true unless dependency.predecessor.state == 'completed'
+    end
+    false
+  end
+end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -75,7 +75,9 @@ class ApplicationController < ActionController::Base
     if todos_parent.nil?
       count = 0
     elsif (todos_parent.is_a?(Project) && todos_parent.hidden?)
-      count = @project_project_hidden_todo_counts[todos_parent.id]
+      count = @project_hidden_todo_counts[todos_parent.id]
+    elsif (todos_parent.is_a?(Context) && todos_parent.hidden?)
+      count = @context_hidden_todo_counts[todos_parent.id]
     else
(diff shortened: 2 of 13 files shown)


dnrce:
Data manipulation in migrations is generally not the best practice. The fact that you're using dummy ActiveRecord classes is smart and this _should_ be fairly durable, but I'd still recommend another approach. I suggest a rake task.

In fact, I think it would be useful to have an "upgrade" task be part of our standard upgrade procedure. We could populate it with any necessary data changes like this, and even have it run the migration step.

dnrce:
(...since you asked specifically for feedback on the migration. I'll look at the rest at a later date -- it looks like a good change it's a bit to digest.)

C-Otto:
Thank you. As additional information, I was inspired by this older migration script: https://github.com/TracksApp/tracks/blob/master/db/migrate/019_convert_todo_to_state_machine.rb

C-Otto:
More comments? What is the number one reason not to merge this in the current state?

C-Otto:
Should todos from hidden projects be shown in the tickler? Currently they are.

C-Otto:
Any comment?

C-Otto:
I've been using this in my installation for two months now, and did not notice any issue. Of course, hidden projects are not part of my daily routine.

db/migrate/20160131233303_remove_project_hidden_state_from_todos.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  class Project < ActiveRecord::Base
+  end
+
+  class Dependency < ActiveRecord::Base
+    belongs_to :predecessor, :foreign_key => 'predecessor_id', :class_name => 'Todo'
+  end
+
+  def self.up
+    @todos = Todo.all
```
mattr-: Using `all` is a bit too expensive. I'd rather see this using something like `find_each` so that we're not loading all the todos into memory of which there could be thousands or hundreds of thousands (for larger installations)

C-Otto: True. That was inspired by a previous script, but I'll update this soon.

C-Otto:
@mattr- Thanks, I re-did the migration script. I'd appreciate any feedback!

db/migrate/20160131233303_remove_project_hidden_state_from_todos.rb
```diff
@@ -0,0 +1,42 @@
...
+        todo.state = 'pending'
+      else
+        todo.state = 'active'
+      end
+      todo.save
+    end
+  end
+
+  def self.down
+    Project.where("state == 'hidden'").find_each do |project|
```
mattr-: you don't have to use strings in `where` calls, you could do:
```
Project.where(state: "hidden").find_each
```

instead.

C-Otto: Thank you, fixed.

mattr-:
Looking good. Let's give @dnrce a day or so to chime in and then if we don't hear anything, I'll go ahead and merge this. :+1:

C-Otto:
Good! I just changed the commit messages.

C-Otto:
@dnrce *poke*

mattr-:
I'll merge this weekend since @dnrce is busy :smiley:






On Fri, Apr 22, 2016 at 12:10 PM -0700, "Carsten Otto" <notifications@github.com> wrote:












@dnrce poke

dnrce:
Don't let me hold you back, @mattr-. :-)

Thanks, @C-Otto!

C-Otto:
This also fixes #1082

mattr-:
:tada: Merged! Thanks @C-Otto!

PR #1983, opened 2016-02-05
quick fix

### LINKED ISSUE #1982: master: Unable to create todos
I'm running the latest master (with #1981). I am able to create ONE todo, but creating a second todo in the same project fails.

Popup: "New Context @Something will be also created. Are you sure?"
Error Message: There was an error retrieving from server: parsererror

Reloading the page or navigating to the project again does not help. Logging out and logging back in helps.

I don't have access to the logs now, will update when necessary (note to self: 14:57:00, project 272).

C-Otto:
fixes #1982

Please have a look at the issue. I'm not sure if this is the right way to go. There may be other similar places. Is this a Google Chrome bug?

### DIFF
--- a/app/assets/javascripts/tracks.js.erb
+++ b/app/assets/javascripts/tracks.js.erb
@@ -201,6 +201,7 @@ var TodoItems = {
         params.data += "&term="+term;
         params.dataType = "json";
         params.async = false;
+        params.cache = false; // fix for Google Chrome around 48.0.2564.82, see issue #1983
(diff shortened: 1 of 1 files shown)


mattr-:
I'm fine working around this particular Chrome bug if it solves your issue, so I'll go ahead and merge it.

dnrce:
Change looks OK but in the future please use (and enforce the use of) meaningful commit messages.

PR #1986, opened 2016-02-09
Don't include a protocol or domain for relative_to_root

mattr-:
Tracks uses this javascript method in the application layout called `relative_to_root` to get relative paths for a lot of Javascript functions. However, it includes the fully qualified protocol and domain for the tracks site in question. If you have Tracks proxied through SSL with nginx, then you get a non-secure URL which causes the Javascript to not work to fetch things like contexts or projects for autocomplete.

Instead of using the `root_url` helper that Rails provides that won't know about SSL, just use a normal forward slash instead, so that you still get a path that's relative to the root instead of a URL.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -15,7 +15,7 @@
       var defaultTags = <%= default_tags_for_autocomplete.html_safe rescue '{}' %>;
       var dateFormat = '<%= date_format_for_date_picker %>';
       var weekStart = '<%= current_user.prefs.week_starts %>';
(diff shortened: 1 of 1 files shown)


mattr-:
Local testing with this makes it seem like it's ok. Going to go ahead and merge this.

dnrce:
Doesn't this break subdir configuration?

mattr-:
Probably. 😔

C-Otto:
Ah! Shame on you, I spent a day without tracks - and half an hour of debugging - because of this. Grrrr. :-1:

mattr-:
Yes, shame on me. I should have reverted this sooner.

Issue #1944, opened 2015-11-29
Modernize the testing infrastructure


mattr-:
I can appreciate the cucumber suite, but locally, on my machine, it renders Firefox useless for me because the webdriver takes complete control of a firefox window.

Let's work on switching to PhantomJS, perhaps with poltergeist or some other such thing that makes sense for what we're trying to do.

dnrce:
How do you feel about RSpec vs MiniTest these days?

mattr-:
I generally default to RSpec because it feels easier to describe test
scenarios with it and I know it better, so I don't fumble around with it as
much. However, I really don't have a strong opinion either way.

mattr-:
This is moving higher up my priority list since the cucumber suite takes at least 75 minutes to run on my machine if I'm doing other things with the machine, and I didn't even let it finish.

C-Otto:
I let it run in a virtual screen (vncserver). This makes the tests invisible, but of course they are not faster because of this.

dnrce:
Yuck.

The real problem is that we have too many tests using Capybara to catch small edge cases. More of these should be captured in true unit tests.

My vote is to move to RSpec (with Capybara where necessary) across the board, and drop both Cucumber and MiniTest. Obviously that would be a substantial change but I have taken stabs at doing pieces of it and it's not terrible. Thoughts?

mattr-:
I'm not sure what caused the test suite to take 75 minutes on my machine before, but I have it consistently running in 20 to 21 minutes now. That means it's moved down my priority list a bit as well.

I'd be onboard with rspec and capybara across the board.

ZeiP:
This issue seems no longer relevant, closing.

PR #1992, opened 2016-02-13
Start using RSpec (again)

dnrce:
Tracks used to use RSpec for its acceptance-testing scenarios and there has been discussion for a long while about going to RSpec across the board. This puts the infrastructure in place and migrates the first few unit tests as a proof of concept. The real goal is to eventually get rid of Cucumber and have one more-or-less integrated test suite.

Questions:
- Do we really like RSpec enough to commit to a migration, or should we attempt to use Capybara via Minitest? See also #1944.
- Are we OK with tests being split between RSpec and Minitest for the duration of the migration?


mattr-:
>Do we really like RSpec enough to commit to a migration, or should we attempt to use Capybara via Minitest? See also #1944.

I prefer RSpec. I feel like it's more expressive and makes it easier to better describe the software being written.

>Are we OK with tests being split between RSpec and Minitest for the duration of the migration?

Yes

dnrce:
Before this gets merged I should flag a problem I'm now running into. I'm attempting to convert the first Cucumber feature to an RSpec feature, and to do this I'm having to switch to using Database Cleaner truncation instead of transactional fixtures. This is fine for the JS features, but the drawback is that I have to disable transactional fixtures globally, and Database Cleaner's `:transaction` strategy is _way_ slower than transactional fixtures. In our current setup this doesn't cause a performance penalty for the units because it only applies to the features, but in an integrated suite, this penalty is felt pretty badly -- the 10 unit specs converted in this PR jump from taking 0.4s to taking 2.3s. Is there a good answer for this? If there's a way to switch strategy between features and other test types, that would be acceptable, but AFAICT the `use_transactional_fixtures` option is global.

bf4:
just use cucumber then

mattr-:
@bf4 do you mean switch the whole entire test suite to cucumber? i'm a bit confused by your comment.

bf4:
> Before this gets merged I should flag a problem I'm now running into. I'm attempting to convert the first Cucumber feature to an RSpec feature, and to do this I'm having to switch to using Database Cleaner truncation instead of transactional fixtures. 

I've never solved this problem without running the out-of-process specs separately from the transactional specs.  So, we can leave the working cucumber features in there, or convert to RSpec and always run 'unit' specs separately from 'integration' specs.

mattr-:
I'm going to close this one out. In the two years since this was opened, I've come around to minitest and would prefer to keep using it going forward.

PR #1994, opened 2016-02-15
On Travis, restrict Cucumber tests to MySQL builds

dnrce:
This will speed up the overall CI execution time.

It's possible this will miss things, but if a database-specific bug is filed because it isn't captured in our unit tests, it's a bug in the tests.

Ideally this could be limited to model tests only, but things aren't that well separated.


mattr-:
:+1: LGTM

Issue #1979, opened 2016-01-31
Travis: Capybara tests fail with "Element is not clickable"


dnrce:
Since #1976, Capybara tests on Travis have extremely high failure rates with errors such as
```
Element is not clickable at point (67, 195.9166717529297). Other element would receive the click:
<div id="navbar" class="navbar-collapse collapse"></div>
(Selenium::WebDriver::Error::UnknownError)
```

Example job: https://travis-ci.org/TracksApp/tracks/jobs/105949666

I don't see these failures locally.

I have tested forcing Travis to use Firefox 44 and it does not help.

mattr-:
I see these failures locally if my Firefox window isn't large enough.

C-Otto:
See #1962

mattr-:
The prior issues with the dependency tests are unrelated to this particular
issue, I believe. Maybe I should start working on getting the cucumber test
suite on webkit or poltergeist or something like that to see if these
issues go away.

dnrce:
This seems to have appeared due to https://github.com/SeleniumHQ/selenium/issues/1202, but that only exposed the issue by making Selenium act more like a real user. I believe the underlying issue is content being hidden by the bootstrap navbar wrapping poorly (from changes in #1923). See screenshots:

<img width="1405" alt="screen shot 2016-02-13 at 1 45 20 pm" src="https://cloud.githubusercontent.com/assets/1084619/13029371/6bf5f78a-d258-11e5-93d7-a0b8e5ec32d5.png">
<img width="1174" alt="screen shot 2016-02-13 at 1 45 27 pm" src="https://cloud.githubusercontent.com/assets/1084619/13029373/706dd030-d258-11e5-9097-bdb52a3810a4.png">
<img width="1028" alt="screen shot 2016-02-13 at 1 45 34 pm" src="https://cloud.githubusercontent.com/assets/1084619/13029374/74aee6c0-d258-11e5-80e4-04104fa02823.png">
<img width="893" alt="screen shot 2016-02-13 at 1 45 39 pm" src="https://cloud.githubusercontent.com/assets/1084619/13029376/78696772-d258-11e5-9c3e-326756d3c5d9.png">

dnrce:
The "old" navbar, by contrast, didn't start to collapse until it got down to around 572px in width.

C-Otto:
Related to #1952 and #1949.

dnrce:
This is primarily a symptom of #1952 and is not a Travis issue.

PR #1996, opened 2016-02-16
Address some Cucumber failures

dnrce:
On Travis, before: 233 failures
After: still 233 failures... but trust me, this will come into its own once #1979 is addressed. :wink:

Locally, for example, this takes dependencies.feature from 10 failures to 0.


mattr-:
Thanks for digging into these!

PR #1998, opened 2016-02-16
Avoid using %i, which requires Ruby 2.0

### LINKED ISSUE #1997: Error in Ruby 1.9.3 with updated preferences page
The `PreferencesControllerTest#test_index_sets_prefs_and_user` is failing in Ruby 1.9.3 only, as of 7652791451959a6433a1906a215d5600c4db30f3 (#1974).

```
$ bin/rake test test/controllers/preferences_controller_test.rb test_index_sets_prefs_and_user
Run options: -n test_index_sets_prefs_and_user --seed 19003

# Running:

E

Finished in 0.655737s, 1.5250 runs/s, 0.0000 assertions/s.

  1) Error:
PreferencesControllerTest#test_index_sets_prefs_and_user:
SyntaxError: ~/src/tracks/app/views/preferences/_date_and_time.html.erb:7: unknown type of %string
'.freeze;   %i{default short long longer}.each do |format|
               ^
~/src/tracks/app/views/preferences/_date_and_time.html.erb:7: syntax error, unexpected $end, expecting keyword_end
'.freeze;   %i{default short long longer}.each do |format|
               ^
    app/views/preferences/index.html.erb:28:in `block in _app_views_preferences_index_html_erb__1336127631248805532_70356743837640'
    app/views/preferences/index.html.erb:10:in `_app_views_preferences_index_html_erb__1336127631248805532_70356743837640'
    test/controllers/preferences_controller_test.rb:29:in `block in <class:PreferencesControllerTest>'

1 runs, 0 assertions, 0 failures, 1 errors, 0 skips
```

dnrce:
Fixes #1997

This should turn the 1.9.3 "lite" (i.e. SQLite) build green.

### DIFF
--- a/app/views/preferences/_date_and_time.html.erb
+++ b/app/views/preferences/_date_and_time.html.erb
@@ -4,7 +4,7 @@
 <br/>
 Or pick one of the following:<br/>
 <div class="form-group btn-group" role="group" data-toggle="buttons">
-  <% %i{default short long longer}.each do |format| %>
+  <% [:default, :short, :long, :longer].each do |format| %>
     <label class="btn btn-default">
       <%= radio_button_tag("date_picker1", t("date.formats.#{format}")) %>
       <%= l(Time.zone.today, format: format) %>
@@ -18,7 +18,7 @@ Or pick one of the following:<br/>
 <br/>
 Or pick one of the following:<br/>
 <div class="form-group btn-group" role="group" data-toggle="buttons">
-  <% %i{default short long longer}.each do |format| %>
+  <% [:default, :short, :long, :longer].each do |format| %>
     <label class="btn btn-default">
       <%= radio_button_tag("date_picker2", t("date.formats.#{format}")) %>
       <%= l(Time.zone.today, format: format) %>


mattr-:
Any reason why we still want to support anything that's not 2.1 or later?

dnrce:
Because our current version of Rails does. If this is the only thing, why wouldn't we? I'll be happy to drop <2.2 support once we're on Rails 5.

mattr-:
Ruby 1.9.3 has been end-of-lifed and is no longer supported. I don't agree with supporting a Ruby version that has reached EOL just because Rails does. Can we drop support for 1.9.3 on master and only support Ruby >= 2.0?

C-Otto:
Debian Oldstable (Wheezy) has 1.9.3 as far as I know. I'm running Stable (Jessie), but I don't know who still uses Oldstable.

dnrce:
I agree that supporting old Rubies isn't worth effort, but AFAIK so far that "effort" is literally only this PR. I can't think of a single instance where use of 1.9.3 or 2.0.0 has been the cause of a help question, either.

mattr-:
:+1:

PR #1999, opened 2016-02-17
Include datepicker i18n only as needed

dnrce:
* Restore the `javascript_tag_for_i18n_datepicker` helper (and actually use it)
* Include the datepicker i18n JS in the asset precompilation list

Ref #1975


mattr-:
:+1:

PR #2010, opened 2016-05-18
Fix recurring todo dialogs

dnrce:
* Position the dialog below the navbar
* Make sure the edit form is placed inside `.legacy-ui` in the DOM

### DIFF
--- a/app/assets/javascripts/tracks.js.erb
+++ b/app/assets/javascripts/tracks.js.erb
@@ -866,7 +866,8 @@ var RecurringTodosPage = {
           RecurringTodosPage.hide_all_recurring();
           RecurringTodosPage.reset_radio();
           $('#recurring_daily').show();
-        }
+        },
+        appendTo: "#content"
       });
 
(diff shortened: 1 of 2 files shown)


C-Otto:
Looks good, although I see a big gray box between the last item ("set the date") and the buttons ("update", "cancel").

dnrce:
True, but it's present in 2.3.

dnrce:
Well, it's not gray in 2.3, it's yellow. But the spacing is unchanged.

dnrce:
My goal in this PR is only to fix the real usability problems. I expect this form to get some updates as we continue to work through the UI redesign.

dnrce:
That all being said, I just added a CSS rule to remove the gray background. It doesn't fix the spacing, but does it make it less obnoxious?

C-Otto:
Yeah, that's better. But I agree, some other issues need to be tackled with higher priority :)

mattr-:
:tada: good stuff

PR #2014, opened 2016-05-26
Remove activerecord-deprecated_finders

### LINKED ISSUE #1916: Clean up deprecated finders
Tracks depends on the `activerecord-deprecated_finders` gem, but this is transitional and things need to be migrated.

* https://github.com/rails/activerecord-deprecated_finders
* http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0-active-record

dnrce:
Fix #1916

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,6 @@ group :development do
   gem "bullet"
   gem "rack-mini-profiler"
   gem "rack-dev-mark"
-  gem "activerecord-deprecated_finders"
 end
 
 group :development, :test do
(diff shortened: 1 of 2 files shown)


dnrce:
Don't merge -- being lazy and letting Travis run the tests for me.

For some reason this is only here as a development dependency so I'm hoping (without actually having done a search) that it can be ripped out without any fighting.

dnrce:
I don't see any regressions in the tests.

This was added at the same time as a Tolk update in b8266bbdd167752052cad40011b997a474e00e46 but Tolk itself doesn't have this gem as a dependency of any sort.

PR #2015, opened 2016-05-29
use @not_done_todos for feeds

### LINKED ISSUE #1726: Per-project/context RSS/iCal feeds show all todos
This is the return of #1684, except text feeds are working. I can reproduce this on trunk by creating three tasks with three different projects, and requesting the feed for one project.

C-Otto:
fixes #1726

the name @not_done_todos is wrong and should be replaced in a dedicated fix

### DIFF
--- a/app/views/todos/index.atom.builder
+++ b/app/views/todos/index.atom.builder
@@ -1,9 +1,9 @@
 atom_feed do |feed|
   feed.title(@feed_title)
   feed.subtitle(@feed_description)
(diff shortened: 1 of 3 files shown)


dnrce:
Thanks! I guess there are no tests for these feeds? It would be nice to add coverage before closing the bug.

dnrce:
I'll move my comments to #2016 -- I'm going to close this one because if we merge into 2.3 we'll want to merge from there into master rather than merging a second PR.

PR #2016, opened 2016-05-29
use @not_done_todos for feeds

### LINKED ISSUE #1726: Per-project/context RSS/iCal feeds show all todos

C-Otto:
fixes #1726

the name @not_done_todos is wrong and should be replaced in a dedicated fix

**Note that this targets the 2.3 branch**

### DIFF
--- a/app/views/todos/index.atom.builder
+++ b/app/views/todos/index.atom.builder
@@ -1,9 +1,9 @@
 atom_feed do |feed|
   feed.title(@feed_title)
   feed.subtitle(@feed_description)
-  feed.updated(@todos.last.updated_at)
+  feed.updated(@not_done_todos.last.updated_at)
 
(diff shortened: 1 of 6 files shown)


dnrce:
Thanks! I guess there are no tests for these feeds? It would be nice to add coverage before closing the bug.

C-Otto:
I added some tests. Feel free to teach me how to get rid of redundancy here, parameterized tests would be nice.

C-Otto:
I guess the CI tests were failing even before these changes?

mattr-:
:boom: thanks for fixing!

Sorry for taking so long to get this merged.

PR #2021, opened 2016-06-06
Convert search page to bootstrap

rbndickson:
How is this?

I removed the JS to focus the cursor in favour of HTML5 autofocus. Should I add a fallback for when it is not supported? 

Browser support info: http://caniuse.com/#feat=autofocus

### DIFF
--- a/app/assets/javascripts/tracks.js.erb
+++ b/app/assets/javascripts/tracks.js.erb
@@ -894,12 +894,6 @@ var RecurringTodosPage = {
     }
 };
 
-var SearchPage = {
-    setup_behavior: function() {
-        $('#search-form #search').focus();
-    }
-};
-
 /**************************************/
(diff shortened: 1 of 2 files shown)


mattr-:
I'd like to see the JS to autofocus continue to work, since Safari doesn't support autofocus on iOS, despite it being supported in iOS's webviews.

rbndickson:
👍 The current search page doesn't autofocus for me on iOS, I don't know if that is just my setup?

dnrce:
Safari on iOS _intentionally_ doesn't honor autofocus. Do we need to work around that?

mattr-:
hmm. the more I think about it, the more I think that we shouldn't really be working around that, so I think this is good as is.

mattr-:
Merging, as the test failures are unrelated to this change.

Issue #2025, opened 2016-06-23
master: UI popups are transparent


C-Otto:
When editing the "show from" of some action, the shown calendar is transparent and the numbers (days) are shown on top of whatever is behind the popup.

EDIT: Same for the context suggestion popup.

I guess this is related to #2024

PR #2026, opened 2016-06-25
Load jquery-ui-rails CSS via "require" instead of "import"

### LINKED ISSUE #2024: master: "Add a new recurring action" UI messed up
In a somewhat recent commit (Rails 4.2?) the UI degenerated as shown in the picture.

recurring

dnrce:
Fixes #2024 and #2025


C-Otto:
Confirmed, thank you!

PR #2055, opened 2017-01-20
make drag&drop todos from/to contexts work again  - fix #2054

### LINKED ISSUE #2054: drag & drop between contexts  broken
tried drag & drop functionality between contexts on version 2.3.0 & on master-branch.
From looking at the code it seems that this has been working at some point in time, but isn't now.
Seems that the "drop_target"'s are missing

scriptythekid:
fix #2054

### DIFF
--- a/app/controllers/todos_controller.rb
+++ b/app/controllers/todos_controller.rb
@@ -397,7 +397,7 @@ class TodosController < ApplicationController
     @todo = current_user.todos.find(params[:id])
     @original_item_context_id = @todo.context_id
     @original_item = current_user.todos.build(@todo.attributes)  # create a (unsaved) copy of the original todo
-    @context = current_user.contexts.find(params[:todo][:context_id])
+    @context = current_user.contexts.find(params[:todo][:context_id].sub! 'c', '')
     @todo.context = @context
     @saved = @todo.save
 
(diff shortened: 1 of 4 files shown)


hirocaster:
LGTM

app/controllers/todos_controller.rb
```diff
@@ -397,7 +397,7 @@ def change_context
     @todo = current_user.todos.find(params[:id])
     @original_item_context_id = @todo.context_id
     @original_item = current_user.todos.build(@todo.attributes)  # create a (unsaved) copy of the original todo
-    @context = current_user.contexts.find(params[:todo][:context_id])
+    @context = current_user.contexts.find(params[:todo][:context_id].sub! 'c', '')
```
mattr-: why would there be a 'c' in the context id from the form?

scriptythekid: no clue, they were present in my vanilla install, and were preventing drag & drop from working.

ZeiP:
The 'c' prefix was caused by the way id was specified in _context.html.erb. Refactored the context elements to use the same underscore format that todos use instead to make it less hacky.

ZeiP:
Finally merged the change. Thanks, @scriptythekid!

Issue #2031, opened 2016-07-09
Upgrade to Rails 5.0


dnrce:
See http://guides.rubyonrails.org/upgrading_ruby_on_rails.html.

Currently blocked by cucumber/cucumber-rails#319.

mathieujobin:
cucumber-rails v1.4.4 released with rails 5.0.0 support

dnrce:
Thanks, @mathieujobin!

dnrce:
I'm currently one failing test away from having a PR for this.

mattr-:
Closed with #2121 instead of #2084.

PR #2060, opened 2017-02-12
Hold rubocop back to fix tests on 1.9.3

dnrce:
We'll git rid of 1.9.3 support with Rails 5 (#2031) but until that happens I'd like the tests to stay green.


dnrce:
Travis looks good. Merging.

PR #2065, opened 2017-02-28
detect links before replacing characters with RedCloth

### LINKED ISSUE #2056: Characters in URL are replaced
I created an action with the note "https://www.youtube.com/playlist?list=PLhuMOCWn4P9io8gtd6JSlI9--q7Gw3epW". The "--" was replaced by a longer version of "-" when shown in tracks (which is OKish), but the resulting link becomes "https://www.youtube.com/playlist?list=PLhuMOCWn4P9io8gtd6JSlI9%E2%80%94q7Gw3epW" which does not work anymore.

C-Otto:
this fixes #2056

Note: this removes a onenote test and might break onenote links!

### DIFF
--- a/test/controllers/todos_controller_test.rb
+++ b/test/controllers/todos_controller_test.rb
@@ -989,6 +989,17 @@ class TodosControllerTest < ActionController::TestCase
     assert_select "a[href='#{url}']"
   end
 
+  def test_url_with_double_hyphen_is_parsed_correctly
+    # issue #2056
+    login_as(:admin_user)
+    todo = users(:admin_user).todos.first
+    url = 'http://example.com/foo--bar'
+    todo.notes = "foo #{url} bar"
+    todo.save!
+    get :index
+    assert_select "a[href='#{url}']"
+  end
+
   def test_link_opened_in_new_window
     # issue #1747
     login_as(:admin_user)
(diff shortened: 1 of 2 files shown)


dnrce:
Unfortunately it's not possible to reverse the order. Autolinking converts URLs in Textile format to HTML, and breaks the textile. For example, a note `"foo":http://foo.com/` is valid Textile but would be converted to `"foo":<a href="http://foo.com/">http://foo.com/</a>` which is no longer valid.

test/controllers/todos_controller_test.rb
```diff
@@ -989,6 +989,17 @@ def test_url_with_slash_in_query_string_are_parsed_correctly
     assert_select "a[href='#{url}']"
   end
 
+  def test_url_with_double_hyphen_is_parsed_correctly
```
dnrce: FWIW I'd prefer to see this as a unit test for `rendering_helper`.

dnrce:
Closing, as this breaks Textile and OneNote support. Will gladly accept an alternative approach.

test/controllers/todos_controller_test.rb
C-Otto: Should I move the other tests too? I did not know about `rendering_helper` when writing the test, by the way.

dnrce: In general we have way too many tests at the wrong level or in the wrong spot. We should aim to true that up as we refactor. For the moment it's a moot point here, though.

C-Otto: Is that a yes or a no? Where's the moot point? I'm confused.

dnrce: I answered generally because I wasn't totally clear what your proposed action is. If you mean open a new PR to move the existing tests, sure, that's good refactoring work. If you mean move the test in this PR, that's a moot point because this PR is closed.

C-Otto: Got it.

PR #2067, opened 2017-03-05
Move and refactor tests for `rendering_helper`

C-Otto:
as discussed in PR #2065

### DIFF
--- a/app/helpers/rendering_helper.rb
+++ b/app/helpers/rendering_helper.rb
@@ -10,7 +10,7 @@ module RenderingHelper
       right = $'
       # detect already linked URLs and URLs in the middle of a tag
       if left =~ /<[^>]+$/ && right =~ /^[^>]*>/
-        # do not change string; URL is alreay linked
+        # do not change string; URL is already linked
         href
       else
         content_tag(:a, h(href), :href => h(href))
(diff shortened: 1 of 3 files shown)


C-Otto:
This now also checks for textile links. Applying PR #2067 now triggers a test failure:

```
Minitest::Assertion: --- expected
+++ actual
@@ -1 +1 @@
-"<p>A link to <a href=\"http://github.com/\">GitHub</a>.</p>"
+"<p>A link to âGitHubâ:<a target=\"_blank\" href=\"http://github.com/\">http://github.com/</a>.</p>"
```

PS: Is that an encoding error?

dnrce:
Great! So much less overhead.

.gitignore
```diff
@@ -26,3 +26,4 @@ db/data.yml
 nbproject
 rerun.txt
 tags
+*.iml
```
dnrce: Can you add this to your system ignore rules instead of the project's `.gitignore`? Git is prescriptive on this point: https://git-scm.com/docs/gitignorehttps://git-scm.com/docs/gitignore

C-Otto: OK, fixed.

dnrce:
Thanks, @C-Otto!

PR #2082, opened 2017-05-20
Move alternative installation options to the wiki

dnrce:
I don't love having pointers to specific third parties in the actual Tracks source, so this change moves all that to the wiki.


mattr-:
:+1: fine by me.

PR #2083, opened 2017-05-20
Add Ruby 2.3 and 2.4 as CI build targets

mattr-:
2.3 and 2.4 have been out for some time. We should probably make sure the project works with them. :smiley:


mattr-:
And the builds fail, so killing this PR for now.

PR #2095, opened 2017-12-23
Resolves a bug (2075) where the project name was encoded with html en…

RyanTruran:
This solves issue 2075. root cause of the problem is that the html entities are stored in the database, for good reasons, however when displaying this data at later times the html entities need to be decoded when html will not do the decoding automatically.

### DIFF
--- a/app/models/stats/projects.rb
+++ b/app/models/stats/projects.rb
@@ -22,7 +22,7 @@ module Stats
 
     def find_top10_longest_running_projects
       projects = user.projects.order('created_at ASC')
-      projects.sort_by{ |p| p.running_time }.take(10)
+      projects.sort_by{ |p| p.running_time }.reverse.take(10)
     end
 
   end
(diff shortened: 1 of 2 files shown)


mattr-:
Could you explain how the jeditable update comes into play here? It looks like just the call to `html_safe` solves the issue.

RyanTruran:
The html_safe only works when putting text in html elements. So it works
well for the project form on the right side.

When I click on the project name to edit it. You will see that html safe
does not fix the text loaded in because it’s not actually modifying the
value in the text field. So the only way to solve this is to edit the
jeditable.js file.

Thanks,
Ryan Truran

mattr-:
The issue with changing the jeditable.js file is that it isn't something we wrote. It's a jQuery plugin we got from elsewhere. If we want to upgrade the jeditable plugin, we'll lose our fix here. It looks like the [jeditable plugin](https://appelsiini.net/projects/jeditable/) documents how to handle this particular situation. Would you mind trying to create a solution that doesn't involve editing the third party plugin?

RyanTruran:
Thanks for the feedback, I modified tracks.js.erb to accomplish the same
thing. with that being said how do i go about pushing this change since i
have already commited and made a pull request for the one that modifies the
jeditable plugin.

PR #2096, opened 2017-12-25
A more rational way to load the text input with the previous names

RyanTruran:
I moved the changes that I originally made on the jeditable plugin to the track.js.erb file.

### DIFF
--- a/app/assets/javascripts/tracks.js.erb
+++ b/app/assets/javascripts/tracks.js.erb
@@ -527,6 +527,10 @@ var ProjectListPage = {
     setup_behavior: function() {
         /* in-place edit of project name */
         $('span#project_name').editable(ProjectListPage.save_project_name, {
+            data:function(encoded, config){
+              var decoded = $('<p></p>').html(encoded).text();
+              return decoded; 
+            },
             style: 'padding: 0px; width=100%;',
             submit: i18n['common.ok'],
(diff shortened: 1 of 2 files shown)


mattr-:
Cool! Thanks!

This is looking good to me. Do you mind if I squash this down into a single commit when I merge it?

RyanTruran:
Is their some way that I can do that?

mattr-:
The way that I do it is in cases like yours would be an interactive rebase (see the section `Squashing commits` from [the Git book](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)) with a force push (`git push -f`) to your repo. GitHub will automatically update the PR here once you've done both of those things.

RyanTruran:
Awesome thanks.

RyanTruran:
I may have messed it up. Do you mind looking at it.

mattr-:
The last commit looks fine but the first two commits aren't needed.  Any chance that you could remove those? We should be good after that.

RyanTruran:
Is that through rebase / drop

mattr-:
> Is that through rebase / drop

Yup. 😃

RyanTruran:
Done

mattr-:
:thumbsup:

mattr-:
Thanks for this contribution!! 

bitmoji

Issue #1745, opened 2015-01-17
Mobile header renders incorrectly in German locale


kralo:
With the german locale, in mobile view on i.e. 360x640 in firefox screen size testing (desktop), but on mobile firefox as well, the "new" button has two lines of text "Neue Aufgabe" and thus is higher than its counterparts.

Because the div #content uses a 'fixed' spacing of top margin 4em, sometimes the navigation bar overlays the content, which is quite ugly. Could the content div not be made floating?

dnrce:
The #content div cannot (trivially) be made floating because the navigation header has a fixed position.

C-Otto:
Maybe just use "+Aufgabe" or similar?

PR #2098, opened 2018-01-23
Fix german locale mobile Header; refs #1745

hewo:
This fixes issue #1745 by shorten link text. The meaning of the text is still clear.


mattr-:
Sorry for taking so long to get to this! 😞 Thanks for the update!

PR #2103, opened 2018-05-02
Update the docker configuration

mattr-:
This moves the Passenger based Dockerfile to its own file and sets up a new Dockerfile that's more suitable for local development and incorporates docker-compose to isolate our database dependencies and any other external services we may want to add in the future.

It also adjusts all the binstubs to use the dockerized environment and using those is now the preferred method for running commands in this repository.

Hopefully this makes it easier for folks to get started hacking Tracks.


mattr-:
This is at the point now where it's running the `lite` CI suite, which is really just the unit tests and none of the cucumber tests. I'm getting the same number of errors locally that I am in CI, which gives me the confidence that this is working properly and as I fix tests locally, that will be reflected in CI without a bunch of flakiness. Merging.

PR #2109, opened 2018-06-20
Upgrade lots of gems

mattr-:
This is pretty risky, as there are some major version bumps in here and I didn't check everything to ensure things are still working. I only ran the test suite. So, YOLO, I guess. ¯\\\_(ツ)\_/¯


greysteil:
@mattr- - would you be up for automating creating gem update PRs? I built [Dependabot](https://dependabot.com) to do that, and it's free for open source. I ran it against a fork and got [these PRs](https://github.com/greysteil/tracks/pulls) - obviously the ones with major version bumps would need work (or can be ignored) but the others look pretty safe.

Thanks for your work on Tracks!

mattr-:
I'd love to do that later! Right now, even some of the minor version bumps give me goosebumps a bit. 

How does Dependabot handle long running branches for stable releases?

greysteil:
Sweet! You can set Dependabot to open PRs against a specific branch.

(If I'm getting the wrong end of the stick and you mean "can it handle keeping me up-to-date with Rails 4.x releases" then the answer to that is "yes", too - it might create a PR to upgrade you to 5.x initially, but once you tell it to ignore that major version it will do the right thing.)

PR #2111, opened 2018-07-21
Remove some friction when running in development

mattr-:
The docker based development environment is great for abstracting away all the various system details (and hopefully makes things easier to get going on Windows) but unless you know all the various switches to toggle to work with docker, it can be a pain in the neck.

This pull request adds a few extra scripts to bring up a rails console or a rails server. It also adds a bit of functionality that was missing before to get `docker-compose` to publish the service ports so you could actually get to the server that you're running :smile:

### DIFF
--- a/script/console
+++ b/script/console
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+appdir=$(cd $(dirname "$0")/.. && pwd)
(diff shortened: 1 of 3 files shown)


irvken:
I asked around at work and no-one had any idea what a dockerized git hub
repository was and they use both. Honestly does not one have a VM lying
around I can steal?

PR #2112, opened 2018-07-25
remove cucumber, some minor gem cleanup

ericmoon:
Removed cucumber/capybara as it isn't used for CI anymore per the resurrection project.
Cleaned up gems after removing the cucumber gems.  Probably a few still hanging out there that can be removed, but this should be the majority.

Travis CI passed, and was able to run the dev docker environment after the changes.


Gemfile
```diff
@@ -67,18 +66,14 @@ end
 group :test do
   # Pull in the fix for rails-dom-testing issue #42
   # TODO: Remove with Rails 5 and rails-dom-testing 2.x
-  gem 'rails-dom-testing', github: 'rails/rails-dom-testing', ref: 'a64f30514ee65f172c43f42cfd4500b5e11a561a'
+    gem 'rails-dom-testing', :git => 'https://github.com/rails/rails-dom-testing', :ref => 'a64f30514ee65f172c43f42cfd4500b5e11a561a'
```
mattr-: Could you switch this back to 1.9 style hashes?
```ruby
git: 'https://github.com/rails/rails-dom-testing', ref: 'a64f30514ee65f172c43f42cfd4500b5e11a561a'
```

Gemfile.lock
```diff
@@ -51,20 +51,10 @@ GEM
...
-    aruba (0.14.5)
-      childprocess (>= 0.6.3, < 0.10.0)
-      contracts (~> 0.9)
-      cucumber (>= 1.3.19)
-      ffi (~> 1.9.10)
-      rspec-expectations (>= 2.99)
-      thor (~> 0.19)
     ast (2.4.0)
-    autoprefixer-rails (8.6.3)
+    autoprefixer-rails (9.0.0)
```
mattr-: Could you revert `autoprefixer-rails` back to 8.6.3? I'd hate for an incompatibility caused by a major version bump to break the site.

Gemfile.lock
```diff
@@ -254,15 +210,13 @@ GEM
       childprocess (~> 0.5)
       rubyzip (~> 1.0)
       websocket (~> 1.0)
-    simplecov (0.16.1)
-      docile (~> 1.1)
+    simplecov (0.13.0)
```
mattr-: This downgrade of `simplecov` is strange to me. Was this caused by something that I can't see at the moment?

ericmoon: Looks like the codeclimate-test-reporter 1.0.8 has a dependency on simplecov at <=0.13.0:

[rubygems page](https://rubygems.org/gems/codeclimate-test-reporter/versions/1.0.8)

Not sure why they put that restriction on it for 1.0.8, since it wasn't there for 1.0.7.. 
Either way, codeclimate-test-reporter is deprecated, so shouldn't spend too many brain cells on this one.  I'll move codeclimate back to 1.0.7 and then move simplecove back to  0.16.1, So it shouldn't have any diff anymore.

ericmoon:
Think I cleaned up all the Gemfile mess I got myself into.  Everything passing with minimal changes, as requested.  Thanks!

ericmoon:
Also, I'm sure you noticed, but for anyone else reading this, I pulled the IRC notification from the .travis.yml file.  I was getting familiar with the code and CI and felt like I was spamming the channel.  Doesn't seem like the channel is used much, so I just commented out the lines.  Feel free to add them back.

mattr-:
Thanks so much for tackling this!

PR #2113, opened 2018-08-02
Test cleanup

ericmoon:
Just some housekeeping for the CI testing.

Updates gem "factory_girl" to "factory_bot" -> factory_girl is deprecated.

Updates "assert_equal nil" to "assert_nil" for future minitest compatibility See
https://github.com/seattlerb/minitest/issues/666  for more info.

There were two or three tests that had a variable as nil, so I couldn't change those to assert_nil directly.  Instead, I went from:

```ruby
assert_equal a , b
``` 
where "a" might be nil

to now:

```ruby
if a.nil?`
   assert_nil a
else
   assert_equal a,b
end
```

which feels hacky - those couple of tests should probably be updated to avoid having the variables as nil.  I'll take a look later, but I'm just getting familiar with the test code.


ericmoon:
Was the a plan to move from fixtures to factories.  I just started looking at the test code and although factory_bot is in the Gemfile, it's never used.

mattr-:
I think `factory_bot` was part of the cucumber suite. Feel free to remove it.

Regarding fixtures vs. factories, I don't intend to move away from fixtures personally, as I find it just as easy to work with fixtures as factories at this point but I won't reject a pull request if somebody feels strongly enough about it to do the work. :smile:

PR #2162, opened 2018-12-06
Fix Docker setup

matjack1:
This commit fixes the Docker setup and also the CI build.

Also the contributing guide is updated with the new information related to Docker. The `CONTRIBUTING.md` file is used to comply with GitHub guidelines to help other developers to contribute.

Personally I would also delete some pages from the wiki as they are outdated and wrong. In particular:
- https://github.com/TracksApp/tracks/wiki/How-to-contribute
- https://github.com/TracksApp/tracks/wiki/Testing
- https://github.com/TracksApp/tracks/wiki/Setting-up-a-development-environment

Hopefully the `CONTRIBUTING.md` file will cover all the information in these three wiki pages and be up to date.

Probably out of scope for this PR, but you can also consider adding [issues and PR templates](https://blog.github.com/2016-02-17-issue-and-pull-request-templates/) and move all three files in the `.github` folder.


mattr-:
> Probably out of scope for this PR, but you can also consider adding issues and PR templates and move all three files in the .github folder.

Yup. I'm getting there. Need to figure out what to put into the templates first. :smiley:

Dockerfile
```diff
@@ -11,4 +11,6 @@ RUN mkdir /app/log
 
 COPY . /app/
 
+RUN cp /app/config/site.yml.tmpl /app/config/site.yml
```
mattr-: I'm not convinced that copying the `site.yml` template into the Docker image is the right thing to do here. Since `site.yml` contains a bunch of customization that a user would need to do to install Tracks, I'd rather leave it out by default.

docker-compose.yml
```diff
@@ -3,17 +3,18 @@ services:
...
     environment:
-      MYSQL_ALLOW_EMPTY_PASSWORD: '1'
+      MYSQL_ALLOW_EMPTY_PASSWORD: 1
+      MYSQL_DATABASE: ${TRACKS_DB:-tracks}
     volumes:
-      - "db-data:/var/lib/mysql"
+      - db-data:/var/lib/mysql
   web:
     build: .
+    command: rails s
```
mattr-: This is unnecessary, IMO, as I've modified the binstubs to use the docker environment by default, so that if you follow standard rails practices and run `/bin/rails s`, it'll start the containers for you.

Dockerfile
matjack1: True, I'll do that, I was just aiming to have a one command only to start, but it's true what you are saying.

What about copying it only if it's not present?

docker-compose.yml
matjack1: I tried that, I ended up adding it as a command because it's not working for me for whatever reason. Let me try some more to figure out what is not working for me. Anyway I agree with your point, I'll get rid of it

mattr-: Interesting. I don't want you to struggle to get it working with the binstubs in the repo, so if it's not a quick fix, I'd rather just leave it in.

matjack1: @mattr- I think I've found it! Isn't that the `./script/server` meant to be run to start the server and then the `./bin/rails *` as a way to run the console or other scripts that might otherwise conflict with the already open port?

If that is the case I might simply need to change the documentation, right?

mattr-: Oh! Yes! `script/server` is definitely meant to be run to start the server so that we expose the service ports to the outside world. If you could update the documentation, that would be great!

matjack1:
Thanks for the feedback @mattr- :)

I think I've sorted out the things related to your comments.

Just force pushed a couple of typo fixes. Let me know if I can also help with the wiki.

mattr-:
@matjack1 Thanks for the fixes here. Sorry for taking so long to get this merged! 😅

matjack1:
oh no worries glad it was useful @mattr- :)

Issue #2036, opened 2016-07-19
REST API: /calendar.xml return nil-classes


RedNifre:
/calendar.xml returns this:

<?xml version="1.0" encoding="UTF-8"?>
<nil-classes type="array"/>

My calendar is empty. Why does it say nil-classes?

ZeiP:
It seems that any empty XML has ”nil-classes” as the root element. This seems like a bug, but affects all APIs, not just /calendar.xml.

ZeiP:
I found a fix for this. It can be found in pull request #2205

ZeiP:
Fixed.

PR #2205, opened 2019-05-03
#2036: Change empty XML responses to have the correct base element

ZeiP:
Adding the :root option to those XML APIs that return a response when no items are found. The ones returning only a single item should return an error response if the requested item was found, and therefore forcing the root element shouldn't be necessary there.


mattr-:
Thanks!

Issue #1139, opened 2014-06-25
Option to disable form autofocus


dnrce:
_Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/1139_

I have Tracks set as my browser home page. If I open a browser window/tab with the intent of visiting somewhere other than my Tracks installation, the form steals focus from the browser's URL and/or search bars before I can finish typing an address or search term. I would like to disable the autofocus on the main page.

_Originally reported by zoombody on April 4, 2011 at 20:27:27 (+0000)_

PR #2206, opened 2019-05-04
Remove the autofocus from the search fields

### LINKED ISSUE #2156: Set autofocus to new issue description instead of search
Related to #1139?

Usually I use tracks to enter new actions more often than using the search functionality. As such I'd prefer the description field to be focussed, instead of the search input field.

mattr-:
This should set the autofocus to the new todo entry form (which already
was set for autofocus) rather than the search box

Fixes #2156
Related to #1139

### DIFF
--- a/app/views/search/_index.html.erb
+++ b/app/views/search/_index.html.erb
@@ -1,7 +1,7 @@
 <div class="container-fluid">
 <%= form_tag(search_results_path, class: 'navbar-form') do %>
   <div class="input-group">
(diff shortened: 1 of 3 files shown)


mattr-:
@C-Otto It looks like it's working as you intended. Would you be able to give it a try before I merge?

mattr-:
Still working as intended. Going to merge this and will address any feedback later.

C-Otto:
I'm a bit late to the party, sorry. Yes, it works! Thanks a lot!

Issue #2032, opened 2016-07-17
REST API doesn't filter by project


RedNifre:
projects/ID/todos.xml returns all todos instead of only the todos of the specified project.

Contexts have the same problem, see https://github.com/TracksApp/tracks/issues/1931

C-Otto:
Duplicate of #1726 I guess, please have a look at #2016 for a possible fix.

RedNifre:
I interpreted the fix in #2016 to mean this, is this correct?:

projects/ID/todos.xml?limit_to_active_todos=1

Doing it like this hides todos of other projects, but it also limits the response to active todos, obviously. However, as a work-around it's fine for me for now, since I just noticed that all todos of a project are contained in projects/ID.xml

ZeiP:
I added a PR #2210 with a fix for this.

ZeiP:
Fixed.

PR #2210, opened 2019-05-04
#2032: Use context and project filters in API also when not requestin…

ZeiP:
…g only done items

This adds the context_id and project_id filters also to the todos variable which contains all todos. The same conditions are already implemented in UndoneTodosQuery, which is used when the limit_to_active_todos parameter is used.


mattr-:
Thanks! Sorry I missed this before!

Issue #2170, opened 2019-02-02
install instructions - migrations do not work and rails server wants docker-compose


benwilliams:
Following the install instructions, the migrations did not work until I changed these

-class CreateTracksDb < ActiveRecord::Migration
+class CreateTracksDb < ActiveRecord::Migration[5.2]

Then trying to start rails server, it complains about missing docker-compose.

mattr-:
Thanks for opening this. The install instructions are meant for the most current release and aren't updated to work with the `master` branch yet. I'll update them before release.

srbaker:
Heya, I'm running into similar. I have a bunch of hacking I want to do on Tracks, but its dependency on Docker is really putting me off. It's such a beautifully simple application, so if the docker dependency can be made "more optional" it would be lovely.

I'll even do the work, if you let me know how you'd like it done.

mattr-:
@srbaker Are you on a platform that makes Docker harder to use? I was trying to make working on this project easier by using Docker and docker-compose so that things would be more normalized.

I'll approve and merge a PR that looks for a specially named dotfile and skips the docker set up if it's found. As long as I can continue using Docker by default, I'm happy to be accommodating to whatever other folks want. :smiley:

srbaker:
Thanks, I'll look into this.

I don't think I'm on a platform that makes Docker harder to use, it just doesn't mesh with how I would like to work.

ZeiP:
Ah. It seems we didn't update the install instructions to cover Docker, this should probably be done quite soon.

boert:
`touch .skip-docker`
will help

sccarr:
hi, any updates on this?  at the 
`bundle exec rails server -e production`
step, I get
`/Users/me/tracks/bin/../script/docker-environment: line 17: exec: docker-compose: not found`
but I am running Mac Mojave and Docker Compose should be included with Docker for OS.  But when I run 
```
me:tracks me$ docker --version && docker-compose --version
Docker version 19.03.8, build afacb8b
-bash: docker-compose: command not found
```
it makes me think Docker Compose is not installed.  I cannot figure out how to install it because every guidance I find assures me it should already be installed.  Would love to see some instructions on the Trax upgrade that cover Docker. 
thanks!!

tet3:
@sccarr I'm not working on macOS, but on both windows and Linux `docker-compose` is installed separately from docker. Did you try `$ brew install docker-compose`?

mattr-:
[...]

PR #2224, opened 2019-05-13
Add Rails version to the migrations.

srbaker:
Got the complaint mentioned in #2170.

With this, both of the issues mentioned there are addressed.

### DIFF
--- a/db/migrate/001_create_tracks_db.rb
+++ b/db/migrate/001_create_tracks_db.rb
@@ -1,5 +1,5 @@
 # Verision 1.0.3 database
-class CreateTracksDb < ActiveRecord::Migration
+class CreateTracksDb < ActiveRecord::Migration[5.2]
   def self.up
     create_table :contexts do |t|
       t.column :name,           :string,      :null => false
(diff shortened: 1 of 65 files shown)


mattr-:
I'm conflicted about this. On the one hand, this is the right thing to do to fix these migrations. On the other, all of them are so old, I'm not sure it makes sense to keep them. Folks can (and should, really) do a `bin/rake db:schema:load` instead of running migrations at first install. Let me think about it for a bit longer.

srbaker:
The reason I need the migrations is because Rails started adding database-specific stuff to `schema.rb`, so I can't db:schema:load on Postgres or SQLite3, even though the app works fine with both.

Any other ideas?

mattr-:
I was considering only supporting MySQL for Tracks going forward, since there are features that only work properly with MySQL. That is one route we could take.

I think I'll just merge this for now, and then open another issue for the database support discussion.

srbaker:
Thanks!

I'm hoping to launch a service based on this, using Postgres. I'd love to work on those features, well abstracted, so I can use Postgres. I'm looking forward to the discussion either way. :D

C-Otto:
I'm using PostgreSQL and like to keep it that way

Issue #2221, opened 2019-05-12
Improve rack-mini-profiler integration and display


mattr-:
rack-mini-profiler is the tool we use in development that shows up in the upper right hand corner of the page.

Here's a screenshot of the small bits of information it displays:
![image](https://user-images.githubusercontent.com/77174/57586643-431b9480-74be-11e9-830c-ab9f5be17253.png)

When you expand one of those, you get a view like this:
![image](https://user-images.githubusercontent.com/77174/57586649-5f1f3600-74be-11e9-9543-28cdcda01823.png)

While I like the information that rack-mini-profiler displays, I really dislike the way it's integrated on the page. I'd prefer to be able to have a bar across the top of the page that could be toggled on or off, based off the way that [Peek](https://github.com/peek/peek) does it. 

If we can't customize how rack-mini-profiler is displayed, then we can switch to peek.

srbaker:
I'll replace this with Peek. Stay tuned.

srbaker:
I've got a PR to replace this with Peek #2228, but I don't like Peek as much as I like MiniProfiler. So I've also got a PR to move MiniProfiler to the bottom left: #2229.

mattr-:
Closed by #2229

PR #2227, opened 2019-05-13
Single quotes don't need to be escaped here.

srbaker:
This was preventing the app from loading, weirdly.

The error I was getting (in all places where I've made changes) looked like this: ```
I18n::InvalidLocaleData: can not load translations from /Users/srbaker/Projects/TracksApp/tracks/config/locales/en.yml: #<Psych::SyntaxError: (/Users/srbaker/Projects/TracksApp/tracks/config/locales/en.yml): found unknown escape character while parsing a quoted scalar at line 585 column 31>```

I don't know why I'm getting this and others aren't: I suspect I'm on a newer `libyaml` that is more strict.

I've confirmed this with the YAML 1.2 spec, and a few other projects around the web, such as: https://github.com/stympy/faker/pull/1305

### DIFF
--- a/config/locales/cs.yml
+++ b/config/locales/cs.yml
@@ -770,11 +770,11 @@ cs:
     edit_recurring_todo: Upravit opakovaný úkol
     error_completing_todo: Nepodařilo se ukončit / aktivovat opakovaný úkol %{description}
     error_deleting_item: Nepodařilo se smazat položku %{description}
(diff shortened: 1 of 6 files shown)


srbaker:
With this one merged, I'll be able to get on with the real work: I'm going to take #2221 and just replace it with Peek.

mattr-:
LOL, Ruby, why? :rofl:

PR #2228, opened 2019-05-13
Replace mini profiler with peek,.

srbaker:
I can't figure out how to pin it to the top because of the navbar, but my CSS-fu is basically non-existent. Would we like rack-mini-profiler better if we pinned it in the bottom?

It's done, but it doesn't look great. And I'm not convinced it's better.

Thoughts?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,11 @@ group :development do
   gem 'tolk', '~> 3.1.0'
 
   gem "bullet"
+
+  gem "peek"
+  gem "peek-git"
+  gem "peek-rblineprof"
+  gem "peek-performance_bar"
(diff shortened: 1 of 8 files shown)


srbaker:
#2229 as an alternative.

mattr-:
I can work on the CSS-fu necessary to get it stuck to the top. Would you mind posting a screenshot?

srbaker:
Closing this one. There are a _lot_ of issues with Peek in non-development mode, and it makes fixing Peek considerably more effort than I want to spend right now.

(Peek developers expect it to be run in production, and I don't care to be thorough enough to do that yet.)

PR #2231, opened 2019-05-19
Change the Flash charts to Charts.js

ZeiP:
This should show exactly the same charts as the old Flash charts (except for the appearance, of course).


ZeiP:
@mattr-, the Codeclimate issues (length and amount of methods) are the same that have previously existed in the controller file. Should these be fixed before merging or could we merge this in to get the charts working?

mattr-:
Nah, we can ignore those. I'll probably just turn codeclimate off for now.

mattr-:
:tada:

C-Otto:
```
stats:195 Uncaught ReferenceError: Chart is not defined
    at initChart (stats:195)
```

I don't see any graphs on my page

ZeiP:
@C-Otto, weird. I'll try to reproduce this. Thanks for reporting!

Issue #1929, opened 2015-09-22
Broken Feed (Text, All starred, active actions)


thatscoteng:
In Tracks 2.3.0, when going to view-> feeds,
Select the text version of 'All starred, active actions' [/todos.txt?tag=starred&token=xyz]

Then get error:
"""
We're sorry, but something went wrong.

If you are the application owner check the logs for more information.
"""

GregSutcliffe:
I see this too. Logs show:

```
I, [2015-09-22T14:16:26.416812 #412191]  INFO -- : Started GET "/todos.txt?tag=starred&token=a38a8f3195d77f45cdd5b0f1794592986b92567d" for 127.11.70.129 at 2015-09-22 14:16:26 -0400
I, [2015-09-22T14:16:26.420332 #412191]  INFO -- : Processing by TodosController#index as TEXT
I, [2015-09-22T14:16:26.420436 #412191]  INFO -- :   Parameters: {"tag"=>"starred", "token"=>"a38a8f3195d77f45cdd5b0f1794592986b92567d"}
E, [2015-09-22T14:16:27.174904 #412191] ERROR -- : PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "taggings"
LINE 1: ... = 'deferred' OR todos.state = 'pending')))) AND (taggings.t...
                                                             ^
: SELECT "todos".* FROM "todos" INNER JOIN contexts c_hidden ON c_hidden.id = todos.context_id WHERE "todos"."user_id" = $1 AND "todos"."state" = 'active' AND (NOT(todos.state = 'project_hidden' OR (c_hidden.state = 'hidden' AND (todos.state = 'active' OR todos.state = 'deferred' OR todos.state = 'pending')))) AND (taggings.tag_id = 27)  ORDER BY todos.due IS NULL, todos.due ASC, todos.created_at ASC
I, [2015-09-22T14:16:27.177582 #412191]  INFO -- :   Rendered contexts/_context.text.erb (19.4ms)
I, [2015-09-22T14:16:27.177753 #412191]  INFO -- :   Rendered todos/index.text.erb (53.8ms)
I, [2015-09-22T14:16:27.178180 #412191]  INFO -- : Completed 500 Internal Server Error in 758ms (ActiveRecord: 19.0ms)
F, [2015-09-22T14:16:27.181305 #412191] FATAL -- : 
ActionView::Template::Error (PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "taggings"
[...]

PR #2233, opened 2019-05-21
#1929: Fix the actions query breaking when querying actions with a no…

ZeiP:
…n-existent tag

### DIFF
--- a/app/models/todos/undone_todos_query.rb
+++ b/app/models/todos/undone_todos_query.rb
@@ -27,6 +27,9 @@ module Todos
 
       if params[:tag]
         tag = Tag.where(:name => params[:tag]).first
+       if not tag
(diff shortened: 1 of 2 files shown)


app/models/todos/undone_todos_query.rb
```diff
@@ -27,6 +27,9 @@ def query(params)
 
       if params[:tag]
         tag = Tag.where(:name => params[:tag]).first
+       if not tag
```
mattr-: Would you mind flattening this out a bit?
```suggestion
       return [] if !tag
```

The english `and`, `or`, and `not` in Ruby behave differently from `&&`, `||`, and `!` respectively and for comparisons like this, I'd like us to stick to the non-english version.

PR #2255, opened 2019-06-25
#1955: Migrate tags to belong to named users for enhanced privacy.

### LINKED ISSUE #1955: Privacy issue: user can see project tags from other users
Version 2.3.0

Repro:
- Have a user A with some projects and tags
- Create a new user B
- Sign up as user B
- Go to Projects page and start filling the form for adding new project
- Start typing into the field "Default tags (optional)", you will get suggestions of project tags from user A

I guess it's not an intended behavior.

ZeiP:
Add the user relation to tags and migrate existing tags to be owned by the user or users that have todos or recurring todos using them. 

Fixes #1955.

Still missing:
- [x] Migration leaves unused tags without owner
- [x] Does the project default tag get created if it doesn't exist?
- [x] For unused tags, assign all the tags to the single user if only one is present, otherwise remove all.
- [x] Document the migration to the changelog.

### DIFF
--- a/db/migrate/20190618202817_add_user_id_to_tag.rb
+++ b/db/migrate/20190618202817_add_user_id_to_tag.rb
@@ -1,72 +1,79 @@
 class AddUserIdToTag < ActiveRecord::Migration[5.2]
   def self.up
-	  add_column :tags, :user_id, :integer
+    add_column :tags, :user_id, :integer
 
-	  # Find uses of each tag for both Todos and RecurringTodos to
-	  # figure out which users use which tags.
-	  @tags = execute <<-EOQ
-		  SELECT t.id AS tid, tds.user_id AS todo_uid, rt.user_id AS rtodo_uid
-		  FROM tags t
-		  JOIN taggings tgs ON tgs.tag_id = t.id
-		  LEFT OUTER JOIN todos tds
-		   ON tgs.taggable_type = "Todo" AND tds.id = tgs.taggable_id
-		  LEFT OUTER JOIN recurring_todos rt
-		   ON tgs.taggable_type = "RecurringTodo" AND rt.id = tgs.taggable_id
-		  WHERE rt.id IS NOT NULL OR tds.id IS NOT NULL
-		  GROUP BY t.id, tds.user_id, rt.user_id
-	  EOQ
+    # Find uses of each tag for both Todos and RecurringTodos to
+    # figure out which users use which tags.
+    @tags = execute <<-EOQ
+      SELECT t.id AS tid, tds.user_id AS todo_uid, rt.user_id AS rtodo_uid
+      FROM tags t
+      JOIN taggings tgs ON tgs.tag_id = t.id
+      LEFT OUTER JOIN todos tds
+       ON tgs.taggable_type = "Todo" AND tds.id = tgs.taggable_id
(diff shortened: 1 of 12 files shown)


ZeiP:
@mattr-, perhaps you'd like to check this out a bit as code review. It's not working yet, but I'm especially interested in the lib/is_taggable.rb changes and if you have better solutions in mind.

mattr-:
> I'm especially interested in the lib/is_taggable.rb changes and if you have better solutions in mind

Nothing better comes to mind there. I'd like to take another look once you get this working before we merge it.

ZeiP:
Yay, it seems to be working now! Additional testing is very welcome. There are a few things I'd like @mattr- and possibly others to chime in, I'll add them as comments in the code review.

db/migrate/20190618202817_add_user_id_to_tag.rb
```diff
@@ -0,0 +1,79 @@
...
+      extras.times do |n|
+        uid = uids[n+1]
+
+        # Create a duplicate of the tag assigned to the user.
+        new_tag = tag.dup
+        new_tag.user_id = uid
+        new_tag.save!
+
+        # Move all the user's regular todos to the new tag.
+        execute <<-EOQ
```
ZeiP: It's not a very good practice to inject variables inside the SQL statement without quoting, but since a migration is a limited scope and all the values are obtained from the same database, I think it's okay. The main problem was that I failed to find a working way for doing a proper prepared statement with placeholders here.

db/migrate/20190618202817_add_user_id_to_tag.rb
```diff
@@ -0,0 +1,79 @@
...
+          WHERE t.user_id = #{uid} AND ta.tag_id = #{tid}
+        EOQ
+      end
+
+      tag.user_id = uids[0]
+      tag.save!
+    end
+
+    # Set all unowned tags to user ID #1, because they're unused.
+    execute <<-EOQ
```
ZeiP: I'm not sure why these occur, since AFAIK tags should be removed when they're no longer used. However these did occur even in my very basic test installation. The question is, should we remove the unused tags or set their user_id to 1?

db/migrate/20190618202817_add_user_id_to_tag.rb
mattr-: I just found [`exec_update`](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/DatabaseStatements.html#method-i-exec_update) and we can use that for prepared statement type things, rather than using a raw `execute` here.

ZeiP: I tried it, but I failed to figure out a working syntax with more than one variable. I spent some hours trying to figure it out, if someone else knows how it works feel free to change it.

db/migrate/20190618202817_add_user_id_to_tag.rb
ZeiP: Yep, you're correct. I noticed that the starred tag is removed when there are no tagged items, and somehow assumed that goes for all the tags – but indeed it seems that starred is the only one removed.

I think #2 sounds reasonable, so I'll get working on that. Thanks!

ZeiP:
@mattr-, does this look ok to you?

PR #2260, opened 2019-06-27
#2259: Use a Chartjs-ror version with airblade/chartjs-ror#57 applied…

### LINKED ISSUE #2259: Enhance quality of JS charts
Should get rid of the width/height specifications and instead have a container doing maintainAspectRatio to have a crispier look on the stats. This requires either airblade/chartjs-ror#57 or manually adding the container elements in the Tracks end.

ZeiP:
Adds a container around each chart allowing to enable responsivity and get a less fuzzy chart.

The PR in upstream hasn't moved for a month, so quickfixing this by referencing straight to the Github repository. Can be changed back to follow upstream once airblade/chartjs-ror#57 is merged.

Fixes #2259.


ZeiP:
@mattr-, any ideas why this is failing?

mattr-:
The initial failures in CI made no sense, but I'm happy to just run CI at pull request time.

ZeiP:
It seems that the CI checks finally passed after I changed the Git repository link to master instead of a branch (which doesn't seem like it should cause an error, but it did), and when I removed the push CI action (which seemed to cause some of the more recent fails, because one check failed and another one didn't). Dunno why, but for me this seems ok now.

dnrce:
Having removed `push`, the CI is no longer building changes to `master`.

ZeiP:
Should it? As I understand it, all changes to master come through pull requests, which means that each change to master has a CI build done anyway.

PR #2261, opened 2019-06-27
#2242: Fix the JS procedure ran on add form submission by changing th…

### LINKED ISSUE #2242: Adding todo: form not emptied, todo list not updated
I'm on 6ad1e2cf93556a45542421f375178da1531d0e2c. When I create a new todo, it is added without any error, but I only see it after refreshing the page. The create-todo-form still shows the todo data. I expect the new todo to be shown without a reload (it was fading in before?), and I expect the form to be empty/filled with defaults.

ZeiP:
…e context ID to the new format

Fixes #2242

### DIFF
--- a/app/helpers/todos_helper.rb
+++ b/app/helpers/todos_helper.rb
@@ -570,7 +570,7 @@ module TodosHelper
   end
 
   def context_container_id(todo)
-    return "c#{todo.context_id}"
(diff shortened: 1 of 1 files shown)


ZeiP:
For some reason the first test round failed but the second passed. Don't see any reason why, so I think it's ok.

PR #2308, opened 2019-12-07
Restyle the new todo forms

mattr-:
This PR restyles the new todo forms to provide more space for the text boxes and the UI overall.

Before:

New todo form - before restyling
New multiple todo form - before restyling

After:
New todo form - after restyling
New multiple todo from - after restyling

This does make the forms take up more space but this looks way better.


mattr-:
Gonna go ahead and merge this since everything is passing.

PR #2328, opened 2020-01-13
Change "todos.description" from "string" to "text"

### LINKED ISSUE #2274: Constraint mismatch: `todos.description` has max length of 300 in model and 255 in DB.
Hi,

Discovered the following issue when analyzing the tracks app as part of a research project. The field `description` in the object `todos` is limited to max length 300 in the model, but 255 in DB. So if a user creates a description greater than 255 but less than 300, it seems they may encounter a problem.

Model validation:
```
class Todo < ApplicationRecord

  MAX_DESCRIPTION_LENGTH = 300
  ...
  validates_presence_of :description
  validates_length_of :description, :maximum => MAX_DESCRIPTION_LENGTH
  ...
```

DB field, which defaults to 255 for string fields in mysql:
```
MariaDB [tracks]> describe todos;
+-------------------+--------------+------+-----+---------+----------------+
| Field             | Type         | Null | Key | Default | Extra          |
+-------------------+--------------+------+-----+---------+----------------+
| id                | bigint(20)   | NO   | PRI | NULL    | auto_increment |
...
| description       | varchar(255) | NO   |     | NULL    |                |
```

I think the best solution may be to limit the length in the model to 255 (and shouldn't create any  conflicts since the field cannot be greater than length 255 in MySQL). Assuming you agree this is an issue, I can create a pull request if that is helpful. Thanks.

whoisutsav:
DB field "todos.description" was type string, which
defaults to length 255 in MySQL. Model allows field
to be up to 300 characters, which could cause a DB
error, if user saved a description of between 255
and 300 characters.

Fixes #2274

### DIFF
--- a/db/migrate/20200109231555_change_todo_description.rb
+++ b/db/migrate/20200109231555_change_todo_description.rb
@@ -1,0 +1,12 @@
+class ChangeTodoDescription < ActiveRecord::Migration[5.2]
+  def up 
+    change_table :todos do |t|
+      t.change :description, :text
+    end
+  end
+  def down
+    change_table :todos do |t|
+      t.change :description, :string
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_06_18_202817) do
+ActiveRecord::Schema.define(version: 2020_01_09_231555) do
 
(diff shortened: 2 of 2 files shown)


whoisutsav:
Fix for issue #2274

mattr-:
Thanks for the PR! I'll test this on a backup of my production database and if I don't lose any data, then I'll merge it.

mattr-:
Didn't lose anything, so going to consider this migration save. Thanks for contributing this! :heart:

PR #2340, opened 2020-02-05
Fix test_start_from_in_future

jyruzicka:
Now we're in 2020, some of the tests in `recurring_todo_test.rb#test_start_from_in_future` are no longer situated in the future.

I've updated the monthly recurring todos so that their start dates will always be in the future, and confirmed that these tests pass.

The "Every weekday" tests still seem to be passing, even though they're now starting in the past. Not sure what's up with that: perhaps I haven't properly grokked the recurring due date logic fully, or perhaps that's something weird in the logic.


mattr-:
Thanks!

PR #2341, opened 2020-02-08
Use Loofah::HTML5::SafeList where possible

jyruzicka:
Another tiny quality-of-life improvement.

From Loofah 2.3.0, `Loofah::HTML5::WhiteList` is deprecated in favour of `Loofah::HTML5::SafeList`. This commit checks to see if `Loofah::HTML5::SafeList` exists, and uses it when it can.


mattr-:
Thanks!

PR #2371, opened 2020-04-08
Fix contribute link in footer and prefer HTTPS

usmanatron:
The "Contribute" link in the footer returns a 404, so have changed to the most relevant looking page.
Also changed all http links to https after confirming they have valid https endpoints.

Note that I haven't run the rake tests because Ive just changed a couple of links (and haven't yet got everything up and running locally).  do let me know if I need to do this!


mattr-:
Thanks for the contribution! :heart:

PR #2403, opened 2020-06-20
doc: Mention Docker skip option

Werkov:
(I only added point no. 4, the rest of diff is just renumbering.)


mattr-:
Thanks for the documentation update! :heart:

PR #2487, opened 2020-09-19
#2486: Change Gitter to Matrix

### LINKED ISSUE #2486: Publish #tracks on Matrix
https://matrix.org/ and https://element.io/ provide an easy-to-join-but-also-Open mechanism to access #tracks on Freenode, which Gitter doesn't quite manage to equal.

This already works, but it would be a good bit smoother if the Matrix servers had OP in the channel and if the #freenode_#tracks:matrix.org room were officially published on Matrix as #tracks.

This documentation on how to do that may be accurate (though note that it is for KDE's self-hosted Matrix server at kde.org rather than the shared one at matrix.org): https://community.kde.org/Matrix#How_to_Add_Your_Existing_IRC_Channel_as_Room

See https://prometheus.io/community/ (and actually go check out the channel and notice how you see all the history) for a great example of this use.

I think it would then be worth recommending this for chat access and dropping Gitter.

ZeiP:
Fixes #2486.

### DIFF
--- a/app/views/integrations/help.html.erb
+++ b/app/views/integrations/help.html.erb
@@ -8,6 +8,6 @@
 
 <ul>
   <li><a href="http://groups.google.com/group/TracksApp">Mailing list</a></li>
-  <li><a href="https://gitter.im/TracksApp/tracks">Gitter chat</a></li>
   <li><a href="https://webchat.freenode.net/#Tracks">IRC channel #Tracks@FreeNode</li>
+  <li><a href="https://matrix.to/#/#tracks:matrix.org">Matrix chat</a> (bridged to the IRC channel)</li>
 </ul>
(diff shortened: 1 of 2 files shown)


app/views/integrations/help.html.erb
```diff
@@ -8,6 +8,6 @@
 
 <ul>
   <li><a href="http://groups.google.com/group/TracksApp">Mailing list</a></li>
-  <li><a href="https://gitter.im/TracksApp/tracks">Gitter chat</a></li>
   <li><a href="https://webchat.freenode.net/#Tracks">IRC channel #Tracks@FreeNode</li>
+  <li><a href="https://matrix.to/#/#tracks:matrix.org">Matrix chat</a> (bridged to the IRC channel)</li>
```
dnrce, line 12: Since it's ultimately pointing to the same resource should this share the `li` with freenode, like in the README?

ZeiP: Yeah, you're probably right. I first thought Matrix is a bit different from IRC and decided to retain the webchat link also, but I now changed both links to be on the same row for clarity.

PR #2551, opened 2021-01-05
Add testing for locales

ZeiP:
Relates to the problem in #2546.


ZeiP:
Adding only the tests in this PR, the CI should fail.

PR #2628, opened 2021-07-24
Simplify the workflow file

mattr-:
This uses a matrix configuration to set up all the various configurations we're testing tracks with. It also adds testing on ruby 3.0 and testing postgres with all the various ruby versions.


mattr-:
The code climate issue is the same issue that was already present. It's flagged as new because I changed the code. 😞

mattr-:
If this gets merged, the branch protection settings will need to be updated as well.

mattr-:
This is cleaner and (IMO) more supportable, so I'll forcefully merge this and update the branch protection settings if I don't see an objection in this issue in the next 24 hours.

ZeiP:
Sorry for the lag! I've been renovating my new apartment and then moving (still not done, unfortunately :) )  To me the changes seem reasonable. Why / what changes are required in the branch protection settings?

.github/workflows/continuous-integration.yml
```diff
@@ -1,38 +1,23 @@
...
-
-  test27sqlite:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v1
-      - run: bash -x script/cibuild 2.7 sqlite
-
-#      - run: bash -x script/cibuild 2.5 postgres
-#      - run: bash -x script/cibuild 2.7 postgres
+      - run: bash -x script/cibuild ${{ matrix.ruby }} ${{ matrix.db }}
```
ZeiP, line 23: The newline at the end of file is missing here; not a big deal, but GitHub, Git and many editors complain about it so it'd be better to add it back.

ZeiP (changes requested):

.github/workflows/continuous-integration.yml
mattr-: Oh, good call. I'll fix that up.

mattr-:
> Why / what changes are required in the branch protection settings?

The names of the checks are changing, so the old check names won't be reported and the new one are `Automated tests / build (<ruby_version>, <database>)` where `<ruby_version>` and `<database>` are defined in the matrix of the workflow file.

ZeiP:
> > Why / what changes are required in the branch protection settings?
> 
> The names of the checks are changing, so the old check names won't be reported and the new one are `Automated tests / build (<ruby_version>, <database>)` where `<ruby_version>` and `<database>` are defined in the matrix of the workflow file.

Ah, you're right. I merged this, thanks!

Issue #2630, opened 2021-07-31
Upgrading to newer Ruby in Ubuntu 20.04 fails


habernal:
I've been running Tracks on Ubuntu 18.04 LTS for a long time but now I have to upgrage do 20.04. It definitely comes with new Ruby versions (but I'm not a Ruby guy), which breaks Tracks.

The first issue was some missing javascript runtime, so I've added

```plain
gem "mini_racer"
```

to the Gemfile. After installing all gems, I get the following error when starting Tracks:

```plain
bundle exec rake db:migrate RAILS_ENV=production
rake aborted!
TypeError: superclass mismatch for class Command
/var/lib/gems/2.7.0/gems/thor-1.1.0/lib/thor/command.rb:2:in `<class:Thor>'
/var/lib/gems/2.7.0/gems/thor-1.1.0/lib/thor/command.rb:1:in `<top (required)>'
/var/lib/gems/2.7.0/gems/thor-1.1.0/lib/thor/base.rb:1:in `require_relative'
/var/lib/gems/2.7.0/gems/thor-1.1.0/lib/thor/base.rb:1:in `<top (required)>'
/var/lib/gems/2.7.0/gems/thor-1.1.0/lib/thor/group.rb:1:in `require_relative'
/var/lib/gems/2.7.0/gems/thor-1.1.0/lib/thor/group.rb:1:in `<top (required)>'
/var/lib/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:324:in `require'
/var/lib/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:324:in `block in require'
/var/lib/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:291:in `load_dependency'
/var/lib/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:324:in `require'
/var/lib/gems/2.7.0/gems/railties-6.0.4/lib/rails/generators.rb:6:in `<top (required)>'
/var/lib/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:324:in `require'
/var/lib/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:324:in `block in require'
/var/lib/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:291:in `load_dependency'
[...]

PR #2632, opened 2021-07-31
Towards fixing #2630

habernal:
There is already gem 'therubyracer', group: :therubyracer , but it was still failing due to a missing JS engine when installing on Ubuntu 20.04


mattr-:
🤔 if therubyracer needs a runtime, what do you think about adding `mini_racer` to the `:therubyracer` in bundler?

PR #2706, opened 2022-01-03
Update Rails to last 6.1

### LINKED ISSUE #2463: Move from ActionMailer::Base.receive to Action Mailbox in processing inbound email
ActionMailer::Base.receive is deprecated and will be removed in Rails 6.1. Use Action Mailbox to process inbound email.

ZeiP:
Fixes #2463.


ZeiP:
Depends on #2463, which seems like there's something to do. Luckily Rails 6.0 will still be supported until June 1st 2023, so no real hurry here.

PR #2709, opened 2022-01-03
Test for Ruby 3.1

### LINKED ISSUE #2924: Fails to run under Ruby 3.1 (Ubuntu 23.04)
Upgrading to Ubuntu 23.04, which includes and upgrade from Ruby 3.0 to 3.1, results in Tracks not running.

Running `bundler install` to reinstall the Gems for the new version of Ruby fails on installing [therubyracer](https://github.com/rubyjs/therubyracer) with the error:

    /usr/lib/ruby/3.1.0/psych/class_loader.rb:99:in `find': Tried to load unspecified class: Libv8::Location::Vendor (Psych::DisallowedClass)

This is a known issue with therubyracer - see [Issue 467](https://github.com/rubyjs/therubyracer/issues/467)

As therubyracer is no longer maintained, the suggested solution is the change over to [mini_racer](https://github.com/rubyjs/mini_racer) by changing the entry in the Gemfile.

There appears to be a configuration mismatch with two of the mini_racer and libv8-node which I worked around by renaming the libv8-node cache directory from `vendor/cache/ruby/3.1.0/gems/libv8-node-16.10.0.0-x86_64-linux-musl/vendor/v8/x86_64-linux-musl` to `
vendor/cache/ruby/3.1.0/gems/libv8-node-16.10.0.0-x86_64-linux-musl/vendor/v8/x86_64-linux`

I've done this and have Tracks working again under Ubuntu 23.04 and Ruby 3.1.2.

ZeiP:
Fixes #2924.


ZeiP:
Seems that the Psych error might be caused by an old Rails version, so this depends on #2706: https://github.com/rails/rails/issues/49977

PR #2744, opened 2022-02-22
Add a user preference that can be used as a fallback for the sender a…

ZeiP:
…uthentication.


ZeiP:
TODO: Check if it handles correctly a case where the mode is 'to' but no mapping or preferences have been defined.

Issue #1416, opened 2014-06-25
[Nginx + webrick] Autocomplete request to localhost


dnrce:
_Migrated from the original issue at https://www.assembla.com/spaces/tracks-tickets/tickets/1416_

I am using Nginx with "proxy_pass http://127.0.0.1:3000;" and ?webrick? (script/server).

Everything works fine except autocomplete.
Apparently the browser is sending the requests to "Request URL:http://127.0.0.1:3000/contexts.autocomplete?term=A" instead of https://$server_ip/[...] where it would be passed by nginx to localhost.

I saw the "Don't use Ajax for autocomplete" ticket, but is there a fix that would work for me now?

PR #2774, opened 2022-04-29
Allow autocompletion behind proxy

Werkov:
When the application is exposed via proxy, i.e. client visible host
(e.g. example.org:443) is different than Ruby server (e.g.
localhost:3000), autocompletion does not work since the generated URLs
refer to the internal hostname.

The AJAX is constructed with root_url and that can be modified with
default_url_options. So the simple fix just allows specifying customized
default_url_options.

Fixes: #1416


ZeiP:
Sorry for the delay, I'll try to review this ASAP.

PR #2962, opened 2023-10-13
Update the Ruby versions in CI

mattr-:
2.7 is EOL and 3.1 and 3.2 are new.


mattr-:
Gonna close this for now. We need to figure out what to do about the assets

PR #2994, opened 2024-02-07
Update Rubies in CI

mattr-:
Wherein we attempt to upgrade things to work on non-EOL rubies. 😅


mattr-:
Closing in favor of #2709, which I forgot about. 😅

PR #3039, opened 2024-07-18
Update bundle config command, add instructions for PostgreSQL

Pegasus-RPG:


ZeiP:
Excellent, thanks! I made a few changes just today to the Gemfile logic, so I'll update it a bit more in the same PR.

PR #3112, opened 2025-06-28
Upgrade to RoR 7.0

ZeiP:


mattr- (approved):
Looks good to me! Much simpler than  I was expecting. ❤️

ZeiP:
> Looks good to me! Much simpler than I was expecting. ❤️

Let's see if it actually works with the tests too. Hope so, but probably the tests uncover something.

